PR #24, opened 2021-08-31
Ensure multimod prerelease can rewrite versions for indirect module deps

Aneurysm9:
Signed-off-by: Anthony J Mirabella <a9@aneurysm9.com>

### DIFF
--- a/multimod/internal/prerelease/prerelease_test.go
+++ b/multimod/internal/prerelease/prerelease_test.go
@@ -61,6 +61,31 @@ require (
 )
 `),
 		},
+		{
+			name: "1.17 style",
+			input: []byte(`module test
+go 1.17
+
+require (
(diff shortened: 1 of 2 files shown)


multimod/internal/prerelease/prerelease_test.go
```diff
@@ -0,0 +1,75 @@
...
+			name: "indirect",
+			input: []byte(`module test
+go 1.17
+
+require (
+	foo.bar/baz v1.2.3 // indirect
+)
+`),
+			expected: []byte(`module test
+go 1.17
```
bogdandrutu, line 57: Add a test with two `require` groups since this is how go 1.17 will create them.

bogdandrutu (approved):

PR #62, opened 2022-02-18
Add dbotconf utility

MrAlias:
Based on the utility added here: https://github.com/open-telemetry/opentelemetry-go/pull/2613

The plan is to replace that utility with this and do the same for [opentelemetry-go-contrib](https://github.com/open-telemetry/opentelemetry-go-contrib)


MadVikingGod (approved):
👍

Aneurysm9 (approved):

PR #67, opened 2022-03-02
Clean up dependabot utility

MrAlias:
Lock dependency to latest version of go.opentelemetry.io/build-tools to
ensure the FindMods function is there. The previous release built
against a dependency that did not include it.

Fix the label for Go updates to be `go` not `actions`.

Use and indent of two spaces to match existing configs.


dbotconf/go.mod
```diff
@@ -5,7 +5,7 @@ go 1.16
 require (
 	github.com/spf13/cobra v1.3.0
 	github.com/stretchr/testify v1.7.0
-	go.opentelemetry.io/build-tools v0.0.0-20220214221246-ac3bcee21caa
+	go.opentelemetry.io/build-tools v0.0.0-00010101000000-000000000000
```
Aneurysm9, line 8: Shouldn't this depend on an actual (pseudo)version of this module?  The commit where `FindMods` was introduced should now be known and referenceable.

MrAlias: Updated.

Aneurysm9 (approved):

PR #87, opened 2022-03-22
Change command config creation logic

bryan-aguilar:
This PR changes how the command config is created to address a bug where `DefaultRunConfig` was never called. Command config creation is now bundled into `newCommandConfig` where commands are added and a run config is created.

### DIFF
--- a/crosslink/cmd/root.go
+++ b/crosslink/cmd/root.go
@@ -31,7 +31,8 @@ type commandConfig struct {
 	pruneCommand cobra.Command
 }
 
-func (c *commandConfig) buildCommands() {
+func newCommandConfig() *commandConfig {
+	c := new(commandConfig)
 	preRunSetup := func(cmd *cobra.Command, args []string) {
 		c.runConfig.ExcludedPaths = transformExclude(c.excludeFlags)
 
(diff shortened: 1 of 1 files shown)


crosslink/cmd/root.go
```diff
@@ -31,7 +31,8 @@ type commandConfig struct {
 	pruneCommand cobra.Command
 }
 
-func (c *commandConfig) buildCommands() {
+func newCommandConfig() *commandConfig {
+	c := new(commandConfig)
```
Aneurysm9, line 35:
```suggestion
	c := &commandConfig{
		runConfig: cl.DefaultRunConfig(),
	}
```

This way there is no point at which this value is not usable.

crosslink/cmd/root.go
```diff
@@ -80,11 +81,12 @@ func (c *commandConfig) buildCommands() {
 		},
 	}
 	c.rootCommand.AddCommand(&c.pruneCommand)
-
+	c.runConfig = cl.DefaultRunConfig()
```
Aneurysm9, line 84:
```suggestion

```

Moving this to the initialization above makes this unnecessary.

Aneurysm9 (approved):

PR #88, opened 2022-03-22
Set rootPath to cwd

bryan-aguilar:
This PR fixes an issue in crosslink where the default root path was not correctly set if the `--root` flag was not provided. The tests were also adjusted to account for this issue.

### DIFF
--- a/crosslink/cmd/root.go
+++ b/crosslink/cmd/root.go
@@ -35,8 +35,17 @@ func newCommandConfig() *commandConfig {
 	c := &commandConfig{
 		runConfig: cl.DefaultRunConfig(),
 	}
+
 	preRunSetup := func(cmd *cobra.Command, args []string) {
 		c.runConfig.ExcludedPaths = transformExclude(c.excludeFlags)
+
+		if c.runConfig.RootPath == "" {
+			cwd, err := os.Getwd()
+			if err != nil {
+				log.Printf("Could not get current working directory: %e", err)
+			}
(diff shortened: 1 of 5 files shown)


crosslink/cmd/root.go
```diff
@@ -35,9 +35,18 @@ func newCommandConfig() *commandConfig {
...
+
 	preRunSetup := func(cmd *cobra.Command, args []string) {
 		c.runConfig.ExcludedPaths = transformExclude(c.excludeFlags)
 
+		if c.runConfig.RootPath == "" {
+			cwd, err := os.Getwd()
+			if err != nil {
+				log.Printf("Could not get current working directory: %e", err)
+			}
+			c.runConfig.RootPath = cwd
```
Aneurysm9, line 47: Set this as the default value of the flag instead so that it is communicated to users.

bryan-aguilar:
@Aneurysm9 I made changes according to the ReadMe spec for [`--root`](https://github.com/open-telemetry/opentelemetry-go-build-tools/tree/main/crosslink#-root). I also made the changes without setting a default value. I was not sure how to convey a default value that could possibly not exist. It is valid to run crosslink in a directory that does not belong to a `.git` repository. As long as a user provides a root flag then there is no error. 

I was not sure how to convey that because the default would be `nil`. It seemed more user friendly to describe that in the description of the flag and then return an error safely.

Aneurysm9 (approved):

PR #158, opened 2022-10-14
[multimod] Add ability for multimod to push tags automatically 

### LINKED ISSUE #157: Can `multimod tag` also push the tags to the upstream repository?

bryan-aguilar:
Closes #157 

Add the ability for multimod to automatically publish tags by providing the `--push` flag.

### DIFF
--- a/.chloggen/tagPush.yaml
+++ b/.chloggen/tagPush.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. crosslink)
+component: multimod
+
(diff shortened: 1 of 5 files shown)


bryan-aguilar:
The prerequisite steps remain the same when using multimod. The only difference is that this PR would give you the ability to push tags automatically during [this](https://github.com/open-telemetry/opentelemetry-go-build-tools/tree/main/multimod#tag-the-new-release-commit) step. Are you looking to push tags that are not controlled/created by multimod? 

Checking the tags before push is not supported in this PR. If that is a requirement there are two options that jump out at me. 
1. Print tags that are being pushed and ask for confirmation before pushing. 
2. Separate push action into a different command that can be run after `multimod tag`. It would be a sub command that would look like `multimod tag push --remote=foo`.

bogdandrutu:
Let's go with the current behavior for the moment. Thanks .

bogdandrutu (approved):

PR #187, opened 2022-10-31
[chore] samll nits in dbotconf verify

bogdandrutu:


dmitryax (approved):

dbotconf/internal/verify_test.go
```diff
@@ -25,7 +25,7 @@ import (
 
 func TestRunVerifyErrors(t *testing.T) {
 	assert.ErrorIs(t, verify(nil), errNotEnoughArg)
-	assert.ErrorIs(t, verify([]string{"", ""}), errTooManyArg)
+	assert.ErrorContains(t, verify([]string{"", ""}), "only single path argument allowed, received")
```
MrAlias, line 28: This change makes these tests more brittle. Changing the message of the existing errors will have no affect on these tests passing, after this change it will. Given we aren't testing the correctness of the message here, rather the fact that an expected error is returned, this doesn't seem like an advantageous change.

bogdandrutu: It is a bit strange to say the least, to wrap a specific error `return fmt.Errorf("%w, received %v", errTooManyArg, args)` in this way just for being able to test.

MrAlias: I'm not following what you mean by strange. Like its something you haven't seen before?

This seems like something I often see in Go code. For example, the go utility in the standard library defines errors this way:

https://github.com/golang/go/blob/b726b0cadb5102bb718f879bede0e76d1e5f5c34/src/cmd/go/internal/modindex/read.go#L125-L128

and then [uses the base](https://github.com/golang/go/blob/b726b0cadb5102bb718f879bede0e76d1e5f5c34/src/cmd/go/internal/modindex/read.go#L677) or [the complete errors](https://github.com/golang/go/blob/b726b0cadb5102bb718f879bede0e76d1e5f5c34/src/cmd/go/internal/modindex/read.go#L139-L141) to determine computation logic.

I think it is functionally better code to decouple the variable from its value. Otherwise, when the value changes the uses of the variable may also need to change. The current code seems like the simplest way to do this to me, but I'm open to changing things to make it clearer if we can preserve the functionality.

bogdandrutu: I've seen that pattern but that is not what it is here. In the example you have a simple/common wrapping, which I believe is a standard way, but here you have concatenation of errors for testing perspective.

Let me explain deeper:

```golang
// I changed err -> errYaml to easily refer to.
if errYaml := yaml.NewDecoder(f).Decode(&c); err != nil {
	return nil, return nil, fmt.Errorf("%w: %v", errInvalid, errYaml)
}
```

The returned error is not a wrapping of the `errYaml`, which is the "not common" patter, and I cannot check that `errors.Is(retErr, errYaml)`.

bogdandrutu: As per my understanding, the wrapping happens with `%w`, so you are wrapping the `errInvalid` and just print `errYaml`.

MrAlias (approved):

PR #217, opened 2023-01-05
semconvgen: add only flag to proccess only a specified type

### LINKED ISSUE #216: semconv: missing support for "only" flag.
To generate the specification 1.16, we need the `--only` flag.

#### Why?
Previously, all semconv sources have been located in a dedicated path. (e.g. `opentelemetry-specifications/semantic_conventions/traces`). In one of the last releases [exception.yaml](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.16.0/semantic_conventions/exception.yaml) was moved out of the `/traces` sub folder. Since then its no longer possible to generate go sem conv.

#### Error
```
opentelemetry.semconv.model.exceptions.ValidationError: Semantic Convention trace-exception reference `exception.type` but it cannot be found! - @2:5
```

See: https://github.com/open-telemetry/opentelemetry-collector/pull/6713

frzifus:
Closes #216

### DIFF
--- a/semconvgen/generator.go
+++ b/semconvgen/generator.go
@@ -167,7 +167,6 @@ func render(cfg config) error {
 		"-v", fmt.Sprintf("%s:/data", tmpDir),
 		cfg.containerImage,
 		"--yaml-root", path.Join("/data/input/semantic_conventions/", path.Base(cfg.inputPath)),
-		"--only", cfg.onlyType,
 	}
 	if cfg.onlyType != "" {
 		args = append(args, "--only", cfg.onlyType)
--- a/semconvgen/generator.go
+++ b/semconvgen/generator.go
@@ -41,6 +41,7 @@ func main() {
 
 	cfg := config{}
 	flag.StringVarP(&cfg.inputPath, "input", "i", "", "Path to semantic convention definition YAML. Should be a directory in the specification git repository.")
+	flag.StringVarP(&cfg.onlyType, "only", "", "", "Process only semantic conventions of the specified type. {span,resource,event,metric,units,scope}")
 	flag.StringVarP(&cfg.specVersion, "specver", "s", "", "Version of semantic convention to generate. Must be an existing version tag in the specification git repository.")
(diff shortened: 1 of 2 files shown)


semconvgen/generator.go
```diff
@@ -162,13 +164,20 @@ func render(cfg config) error {
 
 	args := []string{
 		"run", "--rm",
-		"-v", fmt.Sprintf("%s:/data", tmpDir),
+		"-v", fmt.Sprintf("%s:/data:z", tmpDir),
```
Aneurysm9, line 167: This seems unrelated.  What does this change will do?  A comment might be helpful.

semconvgen/generator.go
```diff
@@ -162,13 +164,20 @@ func render(cfg config) error {
...
 		cfg.containerImage,
 		"--yaml-root", path.Join("/data/input/semantic_conventions/", path.Base(cfg.inputPath)),
+		"--only", cfg.onlyType,
+	}
+	if cfg.onlyType != "" {
+		args = append(args, "--only", cfg.onlyType)
+	}
```
Aneurysm9, line 174:
```suggestion
	}
	if cfg.onlyType != "" {
		args = append(args, "--only", cfg.onlyType)
	}
```

This would always add `--only` and then sometimes add it again.

frzifus: ups, seems to be a left over from a test run :face_exhaling:

semconvgen/generator.go
frzifus: Enforced selinux policies can cause issues when files get shared with a containerized process. Since file permissions are reflected by this labels[0]. Docker/Podman have the ability to relabel[1] shared files. 

Should I remove this from this PR and change it in another one or add a comment on top? What would you recommend?

0: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/security-enhanced_linux/sect-security-enhanced_linux-working_with_selinux-selinux_contexts_labeling_files
1: https://docs.docker.com/storage/bind-mounts/#configure-the-selinux-label

Aneurysm9: Let's address that in a separate PR.

frzifus: Its removed from this one. :)

Aneurysm9 (approved):

PR #253, opened 2023-02-17
bump minimum go version to 1.19

codeboten:
Some dependencies require go 1.19+ https://github.com/open-telemetry/opentelemetry-go-build-tools/actions/runs/4205777490/jobs/7298348033


MrAlias:
Can we bump this to 1.20 instead? Guessing most devs are working with the latest release of Go, and by keeping the CI system using the latest it helps support that development.

We recently did this for https://github.com/open-telemetry/opentelemetry-go/pull/3733. I guess the difference there is we also have compatibility checks, but I'm not sure that matters too much here.

MrAlias (approved):

PR #275, opened 2023-03-13
Add gocp tool

pellared:
## Why

Fixes https://github.com/open-telemetry/opentelemetry-go-build-tools/issues/276

Towards addressing https://github.com/open-telemetry/opentelemetry-go/issues/3846

Towards fixing https://github.com/open-telemetry/opentelemetry-go/issues/3548

Addresses https://github.com/open-telemetry/opentelemetry-go/pull/3841#discussion_r1126790695

## What

Add `gocpy` tool that can be used for sharing internal code (more in `README.md`).

This tool is currently needed for https://github.com/open-telemetry/opentelemetry-go. 

Prior art (with example usage): https://github.com/open-telemetry/opentelemetry-go/pull/3841

## Other 

Do we want to use it for `internal/repo` and `internal/syncerror`? Personally, I think there is no big need as these APIs looks pretty stable. And AFAIK the tools are bumped together. On the other hand, it may be good to dogfood for the sake of consistency across OTel Go repositories. Because I cannot find a clear answer myself, I can simply create a follow-up PR once this one is merged. It should be easier for the reviewers to decide.


MrAlias (dismissed):
The `gocpy` binary needs to be removed from the git history.

gocp/gocp.go
```diff
@@ -0,0 +1,84 @@
...
+	"errors"
+	"os"
+	"strings"
+)
+
+// copy copies a Go file from src to dst
+// with package renamed to pkg.
+func copy(dest, pkg, src string) error {
+	if dest == "" {
+		return errors.New("gocp: destination filepath must be set")
```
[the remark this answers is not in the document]
pellared: Addressed https://github.com/open-telemetry/opentelemetry-go-build-tools/pull/275/commits/517bcbb283c06a0c30b1d25546aa33f4a3a5b12b

pellared:
> The `gocpy` binary needs to be removed from the git history.

Done.

gocp/gocp.go
```diff
@@ -0,0 +1,84 @@
...
+	defer inFile.Close()
+
+	outFile, err := os.Create(dest) //nolint:gosec // This is a file generation tool that takes filepaths as input and output.
+	if err != nil {
+		return err
+	}
+	defer outFile.Close()
+
+	w := bufio.NewWriter(outFile)
+	if _, err := w.WriteString("// Code generated by gocp. DO NOT EDIT.\n\n"); err != nil {
```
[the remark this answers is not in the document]
MrAlias: As long as the OTel Go license is in the top 3 lines it should still work there:

https://github.com/open-telemetry/opentelemetry-go/blob/b62eb2ca88dfccbffb729f2213a2146ee76dc34e/Makefile#L182

Not sure how the collector handles it though.

gocp/README.md
```diff
@@ -0,0 +1,15 @@
+# gocp
+
+`gocp` copies Go source code.
+It can be used for sharing non-exported,
+common code across multiple Go modules.
+
+It is advised to keep the shared code in a separate internal Go module
+e.g. under `internal/shared`.
+The shared code should be consumed aross modules using `go generate`
```
utezduyar, line 9: aross -> across

gocp/README.md
```diff
@@ -0,0 +1,15 @@
...
+e.g. `//go:generate gocp --src=../../internal/shared/env.go --pkg=jaeger --dest=env_shared.go`
+
+`gocp` has been created as one of the solutions
+to avoid depending on internal packages of different modules.
+Such dependencies could transitively lead to build breaking
```
utezduyar, line 14: breaking -> breaking changes?

pellared: I will change to "build failures" 😉

gocp/README.md
pellared:
```suggestion
The shared code should be consumed across modules using `go generate`
```

gocp/README.md
pellared:
```suggestion
Such dependencies could transitively lead to build failure
```

utezduyar (commented):
LGTM for my review comments

gocp/main.go
```diff
@@ -0,0 +1,48 @@
...
+	-p, --pkg string    The destination package name (can contain comments).
+	-s, --src string    The source filepath.
+*/
+package main
+
+import (
+	"fmt"
+	"os"
+
+	flag "github.com/spf13/pflag"
```
MrAlias, line 31: Why not just use [`flag`](https://pkg.go.dev/flag)?

gocp/gocp.go
```diff
@@ -0,0 +1,84 @@
...
+		return err
+	}
+	scanner := bufio.NewScanner(inFile)
+	pkgReplaced := false // Rename the line with package statement only once.
+	for scanner.Scan() {
+		txt := scanner.Text()
+
+		if !pkgReplaced && strings.HasPrefix(txt, "package ") {
```
MrAlias, line 58: This seems like it has limited functionality compared to using a text template. When these copied internal packages start to depend on each other the imports will need to be appropriately templatized. Why not just use a text template for this?

gocp/main.go
pellared: Only because this package is used in all other tools. Indirectly in most of them b/c of spf13/cobra and directly in `crosslilnk` and `semconvgen`. `flag` is not directly used in any of the tools.

gocp/gocp.go
pellared: > When these copied internal packages start to depend on each other the imports will need to be appropriately templatized.

That is true. Currently, we already have a use case for it: https://github.com/open-telemetry/opentelemetry-go/blob/4ccd59056982f34c3fcc8faef574f209b70d0f59/sdk/internal/env/env_test.go#L24

>  Why not just use a text template for this?

I found it easier to use actual Go code than writing Go source code templates because we get IDE support. 

The second reason is that the the "shared Go code" is analyzed using `golangci-lint`  what we could not say about templates. The problem is that if the generated code has `// Code generated by gocp. DO NOT EDIT.` comment then [it won't be analyzed by `golangci-lint`](https://github.com/golangci/golangci-lint/blob/076f6b911d718cf9845fe338d40972f41510c4fe/pkg/result/processors/autogenerated_exclude.go#L72-L93). We could alter the text to something like `// Generated by gocp.` However, we will lint all the generated files instead of only linting the source files .

**Summary**

Having all that said I feel that we indeed need this flexibility and I will try to simply create a simple CLI for text/template.

pellared:
Closing per https://github.com/open-telemetry/opentelemetry-go-build-tools/pull/275#discussion_r1141751735. I will create a new PR (next iteration) in the following days.

Issue #41, opened 2021-10-21
semconvgen: panic: unable to exe when specification version is set


MrAlias:
```
$ .tools/semconvgen -i ~/code/opentelemetry-specification/semantic_conventions/resource -s "v1.7.0" -t semconv/template.j2
panic: unable to exec /usr/bin/git worktree add /tmp/otel_semconvgen1125124/input v1.7.0: exit status 128

goroutine 1 [running]:
main.main()
	/home/tyler/go/pkg/mod/go.opentelemetry.io/build-tools/semconvgen@v0.0.0-20210920164323-2ceabab23375/generator.go:59 +0x42e
```

Version: `	go.opentelemetry.io/build-tools/semconvgen v0.0.0-20210920164323-2ceabab23375`

pellared:
I think the proper fix is to improve the error handling and tell that `git worktree add` failed. Maybe the command is emitting something to stdout/stderr that is worth printing? I would not like to run `worktree add` with `-f` flag, because it can be dangerous as it will create a "linked worktree" that will modify the existing one.

PR #301, opened 2023-04-22
Force worktree add

unicod3:
Resolves https://github.com/open-telemetry/opentelemetry-go-build-tools/issues/41

I also came upon this issue while playing with the semconvgen package and tried to patch it.

More on the `-f` param:
https://git-scm.com/docs/git-worktree#Documentation/git-worktree.txt---force


MrAlias (approved):
Please add a changelog entry.

pellared (approved):

pellared:
@unicod3 I am not sure it if is needed. I guess I encountered the same issue here: https://github.com/open-telemetry/opentelemetry-go/pull/4078#discussion_r1188555233

I am not sure if `-f` does always what we want:

> [...] if <path> is already assigned to some worktree but is missing (for instance, if <path> was deleted manually). This option overrides these safeguards.

PR #308, opened 2023-05-10
crosslink: Support go.work

pellared:
## Why

Towards https://github.com/open-telemetry/opentelemetry-go/issues/3964

`go.work` is honored by `gopls` language server so that develoeprs get better UX in tools like VSCode, Vim, Emacs etc

References:
- https://go.dev/ref/mod#workspaces
- https://go.dev/blog/get-familiar-with-workspaces
- https://go.dev/doc/tutorial/workspaces

## What

Make `crosslink` to update the `use` in `go.work` file (if it exists).

Take notice that it will not harm any repositories which do not use Go workspaces. However developers could still use the functionality during development 😉 

PS. The tests have a lot of repetition . However refactoring it in this PR would make it very hard to review. I can refactor all crosslink unit tests in next PR. 

## Testing

You can check the result on this branch and on https://github.com/open-telemetry/opentelemetry-go/pull/4083


hanyuancheung (approved):

bryan-aguilar (commented):
Nice! This looks cool, I'm glad you have a use case for this and thanks for contributing it back!

My one question is, should updating a files `go.work` and `go.mod` files always happen at the same time. With the current implementation this seems to imply that is what a user would always want. 

Is there a scenario where you would only want to update `go.mod` files and not `go.work` files? What about the other way around? 

Should users be able to opt in or out of these updates through command flags? Should we instead be separating the updating of a `go.work` file behind a different cobra command?

bryan-aguilar:
Also, should we support the use case of creating a `go.work` file if it does not exist? Why should having a preexisting `go.work` file be a prerequisite?

crosslink/internal/common.go
```diff
@@ -38,6 +41,31 @@ func identifyRootModule(rootPath string) (string, error) {
...
+
+		localPath, err := filepath.Rel(rootModulePath, module)
+		if err != nil {
+			return nil, fmt.Errorf("failed to retrieve relative path: %w", err)
+		}
+		if localPath == "." || localPath == ".." {
+			localPath += "/"
+		} else if !strings.HasPrefix(localPath, "..") {
+			localPath = "./" + localPath
+		}
```
bryan-aguilar, line 62: This logic looks very similar to the logic found in `insertReplace` [here](https://github.com/open-telemetry/opentelemetry-go-build-tools/blob/653d34f660a4223b39ffdf37f4dc81dd419887c9/crosslink/internal/crosslink.go#L75). Can we lift this out and move it into `common.go`?

go.work
bryan-aguilar, line 1: Should we be checking this `go.work` file in? Is this intentional? From what I remember the `go.work` docs had suggested that these were not suitable for checking in, and only existed for local development. Have we identified a use case where that is not true?

pellared:
> Is there a scenario where you would only want to update `go.mod` files and not `go.work` files? What about the other way around?

I was thinking about it as well and I have not come up with such scenario.

pellared:
> Also, should we support the use case of creating a `go.work` file if it does not exist? Why should having a preexisting `go.work` file be a prerequisite?



I was thinking about it as well. For sure it should be an opt-in functionality as somebody may want NOT to use Go workspaces.  I found following reasons why I did not want to add such functionality (at least in this PR):

1. Adding a new `go.work` file is simply done by running `go work init`. I do not think it is worth to make `crosslink` more complex is creating the file is so easy.
3. We would probably need to add a logic for "guessing" a proper Go version. For instance in this repo I had to manually change `go 1.20` to `go 1.19`
4. YAGNI 😉 This can be always added later

crosslink/internal/common.go
pellared: I agree. I will to it tomorrow 😉

pellared: Addressed 27832a32b9403873e246ba99c9022783138b65de

pellared:
> > Is there a scenario where you would only want to update `go.mod` files and not `go.work` files? What about the other way around?
> 
> I was thinking about it as well and I have not come up with such scenario. I do not see why anyone would like to update only `go.mod` without updating `go.work` and vice-versa. Doing it separately may make the repo "out-of-sync".

I see a use case. Someone may only use `go.work` and not care about having `replace` in `go.mod` files 😉 Do you have any proposal how the CLI should look like? Personally, I propose to add flags to ignore processing of `go.mod` or/and `go.work` files. Then by default user can still update both type of files using one command. Also `prune` could still support both type of files.

Personally, I propose to add it as a separate PR to make it easier to review. Also I do not think it is a blocker as this can be always added on top of existing PR.

go.work
pellared: Removing `go.work` because of https://github.com/golang/go/issues/60126.

pellared:
@bryan-aguilar After longer thoughts I am considering to change the code to have a new `gowork` command. Then the command could generate a `go.work` file if it does not exist. What do you think?

bryan-aguilar:
> @bryan-aguilar After longer thoughts I am considering to change the code to have a new `gowork` command. Then the command could generate a `go.work` file if it does not exist. What do you think?

I think this makes sense. My suggestion would have been very similar. I would expect the cli interface to be something like `crosslink gowork`. I am sure there will be some flags that are shared between the `root` and `gowork` command.

go.work
pellared: https://github.com/open-telemetry/opentelemetry-go-build-tools/pull/308/commits/3fe231b4a93a6b0a9858c5fd5ffbbd83e67e4fc5

pellared:
Moving to draft. I will change this PR to introduce a `work` command. ETA Monday/Tuesday.

pellared:
Closing this PR in favor of https://github.com/open-telemetry/opentelemetry-go-build-tools/pull/311

CC: @bryan-aguilar @hanyuancheung

PR #320, opened 2023-05-22
Include /gotmpl in .github/dependabot.yml

CharlieTLe:
This will configure dependabot to update the dependencies for the gotmpl module. It seems like it's already being updated since https://github.com/open-telemetry/opentelemetry-go-build-tools/pull/319 updated it recently. Though, I'm explicitly adding it here to ensure that it gets updated.

cc: @pellared

### DIFF
--- a/.chloggen/dbotconf-gen-gotmpl.yaml
+++ b/.chloggen/dbotconf-gen-gotmpl.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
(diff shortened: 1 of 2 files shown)


MrAlias (approved):

pellared (approved):
@CharlieTLe Nice catch. It looks like there is a bug in `dbconf verify`. The `ci` GitHub workflow should fail here: https://github.com/open-telemetry/opentelemetry-go-build-tools/blob/6fd2bd763e919eb562f800dd9df3d63d6a5dd035/.github/workflows/ci.yml#L41
as `make dependabot-check` should fail.

pellared:
Superseded by https://github.com/open-telemetry/opentelemetry-go-build-tools/pull/323

PR #326, opened 2023-06-02
Offer to use class and type instead of component

atoulme:
This allows to use the `component` field in changelog, or to use the `class` and `type` fields. This optional use will allow us down the road to formalize changelog per component.

## Related issue ##
https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/23249


chloggen/internal/chlog/entry.go
```diff
@@ -82,7 +120,11 @@ func (e Entry) String() string {
 	issueStr := strings.Join(issueStrs, ", ")
 
 	var sb strings.Builder
-	sb.WriteString(fmt.Sprintf("- `%s`: %s (%s)", e.Component, e.Note, issueStr))
+	component := e.Component
```
pellared, line 123: I would rather rename it to something like `prefix`

chloggen/internal/chlog/entry.go
```diff
@@ -47,6 +55,15 @@ var changeTypes = []string{
...
 }
 
+var classTypes = []string{
+	Receiver,
+	Exporter,
+	Processor,
+	Connector,
+	Extension,
+	Cmd,
+}
```
pellared, line 65: This is tightly coupled to Collector. Should it not be configurable?

pellared (commented):
> This optional use will allow us down the road to formalize changelog per component.

It there any issue which describes/discusses the plan/design?

chloggen/internal/chlog/entry.go
atoulme: This is indeed tied to the collector, all other uses can keep using the component field just like today.

chloggen/internal/chlog/entry.go
atoulme: Not quite sure I follow, this is not used as a prefix, it's still used as the component name if provided.

atoulme:
@pellared I have filed https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/23249 to document the plan.

atoulme:
Closing ; will open a new PR with the new approach based on package.

PR #350, opened 2023-07-05
add multimod diff command

codeboten:
This command checks if there have been any changes for a specific module since a specified version. This will be used to replace the shell scripting in `make check-changes` in the collector repository.


mx-psi:
Looks like a test is failing:

```
--- FAIL: TestHasChanged (0.23s)
    --- FAIL: TestHasChanged/changes_found,_tag_exists (0.06s)
        diff_test.go:89: 
            	Error Trace:	/home/runner/work/opentelemetry-go-build-tools/opentelemetry-go-build-tools/multimod/internal/diff/diff_test.go:89
            	Error:      	Received unexpected error:
            	            	tag not found v0.8.0
            	Test:       	TestHasChanged/changes_found,_tag_exists
FAIL
```

pellared:
@codeboten PTAL

https://github.com/open-telemetry/opentelemetry-go-build-tools/blob/56e8a8a3a736e34a90c275b73720f45041faa841/CONTRIBUTING.md#L43-L48

I will try to review this PR in the following days 😉

multimod/internal/diff/diff.go
```diff
@@ -0,0 +1,100 @@
...
+
+	"github.com/go-git/go-git/v5"
+
+	"go.opentelemetry.io/build-tools/multimod/internal/common"
+)
+
+// check
+
+// normalizeVersion ensures the version is prefixed with a `v`.
+func normalizeVersion(ver string) string {
```
pellared, line 20: Why are we doing it?  I find it YAGNI and also it makes the tool not flexible. What if someone wants to compare with some custom tag.

multimod/internal/common/git.go
```diff
@@ -155,3 +156,53 @@ func VerifyWorkingTreeClean(repo *git.Repository) error {
...
+	}
+
+	o, err := r.TagObject(tagRef.Hash())
+	if err != nil {
+		return nil, fmt.Errorf("tag object error %s %w", tagRef.Hash().String(), err)
+	}
+	return r.CommitObject(o.Target)
+}
+
+// FilesChanged returns the differences between two commits
```
pellared, line 189: This comment is inaccurate. It only list files. It does not really say what are the differences.

multimod/internal/diff/diff.go
```diff
@@ -0,0 +1,100 @@
...
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/go-git/go-git/v5"
+
+	"go.opentelemetry.io/build-tools/multimod/internal/common"
+)
+
+// check
```
pellared, line 17: ? 😉

multimod/internal/diff/diff.go
```diff
@@ -0,0 +1,100 @@
...
+	headCommit, err := client.HeadCommit(r)
+	if err != nil {
+		return changed, changedFiles, err
+	}
+
+	// get all modules in modset
+	for _, tagName := range tagNames {
+		// check tag exists
+		tag := normalizeTag(tagName, ver)
+
```
pellared, line 78: nit: this line can be removed

pellared (commented):
@codeboten  Review done 👍 Will you be able to address the comments?

multimod/internal/diff/diff.go
codeboten: This is here because the missing `v` prefix in the version was causing problems in the past in the collector repo. So much so that we ended up implementing this logic in the Makefile. Since this tool was supposed to replace that functionality, i decided to implement it here.

I suppose this could be an additional flag passed into the diff command to allow others to avoid the code if they didn't need it?

pellared: Adding a comment which will tell why it is added is good enough for me 😉

PR #360, opened 2023-07-10
[chloggen] Add new change_type value for library API changes

mx-psi:
Updates open-telemetry/opentelemetry-collector-contrib/issues/24014.

This adds a new `library_api_change` type that would apply to Go API only changes in the OpenTelemetry Collector contrib repository (e.g. those related to open-telemetry/opentelemetry-collector-contrib/issues/17273).

### DIFF
--- a/chloggen/internal/chlog/entry.go
+++ b/chloggen/internal/chlog/entry.go
@@ -24,11 +24,12 @@ import (
 )
 
 const (
-	Breaking     = "breaking"
-	Deprecation  = "deprecation"
-	NewComponent = "new_component"
-	Enhancement  = "enhancement"
-	BugFix       = "bug_fix"
+	Breaking         = "breaking"
+	Deprecation      = "deprecation"
+	NewComponent     = "new_component"
+	Enhancement      = "enhancement"
+	BugFix           = "bug_fix"
+	LibraryAPIChange = "library_api_change"
 )
 
 type Entry struct {
--- a/.chloggen/mx-psi_library-api-change.yaml
+++ b/.chloggen/mx-psi_library-api-change.yaml
@@ -8,7 +8,7 @@ component: chloggen
 note: Add new 'library_api_change' change_type value.
 
 # One or more tracking issues related to the change
-issues: []
+issues: [360]
 
 # (Optional) One or more lines of additional information to render under the primary note.
 # These lines will be padded with 2 spaces and then inserted directly into the document.
--- a/.chloggen/mx-psi_library-api-change.yaml
+++ b/.chloggen/mx-psi_library-api-change.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. crosslink)
+component: chloggen
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add new 'library_api_change' change_type value.
+
+# One or more tracking issues related to the change
+issues: []
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: This is a value specific to the OpenTelemetry Collector Contrib repository to indicate API only changes
--- a/chloggen/internal/chlog/summary.go
+++ b/chloggen/internal/chlog/summary.go
@@ -32,6 +32,7 @@ type summary struct {
 	NewComponents   []string
 	Enhancements    []string
 	BugFixes        []string
+	LibraryChanges  []string
 }
 
 func GenerateSummary(version string, entries []*Entry) (string, error) {
@@ -51,6 +52,8 @@ func GenerateSummary(version string, entries []*Entry) (string, error) {
 			s.Enhancements = append(s.Enhancements, entry.String())
 		case BugFix:
 			s.BugFixes = append(s.BugFixes, entry.String())
+		case LibraryAPIChange:
+			s.LibraryChanges = append(s.LibraryChanges, entry.String())
 		}
 	}
 
@@ -59,6 +62,7 @@ func GenerateSummary(version string, entries []*Entry) (string, error) {
 	s.NewComponents = sort.StringSlice(s.NewComponents)
 	s.Enhancements = sort.StringSlice(s.Enhancements)
 	s.BugFixes = sort.StringSlice(s.BugFixes)
+	s.LibraryChanges = sort.StringSlice(s.LibraryChanges)
 
 	return s.String()
 }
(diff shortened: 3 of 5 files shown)


chloggen/internal/chlog/entry.go
```diff
@@ -24,11 +24,12 @@ import (
...
-	Deprecation  = "deprecation"
-	NewComponent = "new_component"
-	Enhancement  = "enhancement"
-	BugFix       = "bug_fix"
+	Breaking         = "breaking"
+	Deprecation      = "deprecation"
+	NewComponent     = "new_component"
+	Enhancement      = "enhancement"
+	BugFix           = "bug_fix"
+	LibraryAPIChange = "library_api_change"
```
dmitryax, line 32: As I mentioned in https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/24014, this new type can be confusing for the go instrumentation library. All the changes are library changes. If "properly" follow this new type, we combine all of them under one section.

Even for the collector, we need to have a way to differentiate different types of "Go API only" changes. We have a lot of them in core.

Can we maybe have another orthogonal input like a boolean `library_change`? That can be used to generate two sections for library and user-facing changes, each section will have all the types if applicable, like:

```
## 1.0 User-facing changes

### 🛑 Breaking changes 🛑
foo
bar

### 🚩 Deprecations 🚩
foo
bar

### 🚀 New components 🚀
new component

### 🧰 Bug fixes 🧰
foo
bar
foobar


## 1.0 Library changes

### 🛑 Breaking changes 🛑
foo
bar

### 🚩 Deprecations 🚩
foo
bar

### 🚀 New components 🚀
new component

### 🧰 Bug fixes 🧰
foo
bar
foobar
```

djaglowski: > Can we maybe have another orthogonal input like a boolean library_change? 

I agree this is an orthogonal concern, but library api changes are not mutually exclusive with user-facing changes. For example, when adding an option to a configuration we must make an API change as well. So while I agree at a high level, I propose we should have an orthogonal input that supports the three cases of user-only, api-only, or both.

Additionally, in the interest of keeping this tool somewhat generic, I suggest that the input be a slice where the possible values are enumerated (possibly as part of the configuration of the tool). e.g. 

```yaml
# only relevant to users
audience: [ user ]

# only relevant to api users
audience: [ api ]

# relevant to both
audience: [ user, api ] # make it optional and use this as the default?
```

mx-psi: > this new type can be confusing for the go instrumentation library

This is not meant to be used for the Go instrumentation library, or even for the Collector core I'd say, only for Collector contrib. I think any change we make, including `audience` is going to have that problem: what is the distinction between `user` and `api` on opentelemetry-go?

> Even for the collector, we need to have a way to differentiate different types of "Go API only" changes. We have a lot of them in core.

Issue https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/24014 was meant to be about contrib only. If you want to include core, I feel like the discussion is very different.

> I agree this is an orthogonal concern, but library api changes are not mutually exclusive with user-facing changes. For example, when adding an option to a configuration we must make an API change as well. So while I agree at a high level, I propose we should have an orthogonal input that supports the three cases of user-only, api-only, or both.

What kind of changes would be user-only? AIUI, any user-facing change is also a library change, since it changes the behavior of at least one symbol.

> And I can imagine even more options. E.g., if we change mdatagen to generate new internal code, users and api are not affected but users of mdatagen command outside of the repo might need to change a lot in their codebase to adapt to the new generated code. Maybe that can be another option like tooling. In that case, audience probably is not the best word, maybe affects?

Isn't that just a change to the `mdatagen` component? Why would it need to be different?

----

General comment: while I am okay with the changes proposed above, I think they are very different from what I had in mind with Option B on open-telemetry/opentelemetry-collector-contrib/issues/24014 and I would prefer if either of you opens a PR with a concrete proposal in that line and we can close this one if your PR ends up being the accepted solution.

djaglowski: > What kind of changes would be user-only? AIUI, any user-facing change is also a library change, since it changes the behavior of at least one symbol.

Bug fixes often involve no changes to anything exported by the library. Performance enhancements can be made with no api changes.

djaglowski: > they are very different from what I had in mind with Option B

Is it just a different implementation than you had in mind? Option B was "Have a separate changelog section for Go API only changes", which I think is the goal of either approach.

dmitryax: > What kind of changes would be user-only? AIUI, any user-facing change is also a library change, since it changes the behavior of at least one symbol.

Another kind of breaking change for the users is a change to emitted data. It doesn't change the go API in any way.

> Isn't that just a change to the mdatagen component? Why would it need to be different?

I mean, a change can be breaking for those who use the `mdatagen` tool outside of the repo, but it's not a breaking change for Collector end users or Go API.

mx-psi: > Bug fixes often involve no changes to anything exported by the library. Performance enhancements can be made with no api changes. I agree deprecations cannot be only user facing though.

Right, they are a *library change* but not a *library API change*. I was thinking about library changes in general, where I'd say any user-facing change is also a change of this kind, but I agree that the API itself (what symbols exist and what their names and types/signatures are) does not change in those cases.

> Is it just a different implementation than you had in mind? Option B was "Have a separate changelog section for Go API only changes", which I think is the goal of either approach.

I see https://github.com/open-telemetry/opentelemetry-go-build-tools/pull/360#discussion_r1258613553 as Option A, these are two separate changelogs, even if on the same file. Sorry, I feel like the original issue was not clear enough as to what each option entailed.

djaglowski: > Right, they are a _library change_ but not a _library API change_. I was thinking about library changes in general, where I'd say any user-facing change is also a change of this kind, but I agree that the API itself (what symbols exist and what their names and types/signatures are) does not change in those cases.

What's the case for documenting _library changes_ that are not _library API changes_? In my mind, a _library change_ that is not a _library API change_ is unremarkable except in the way that it impacts users.

djaglowski: > I see https://github.com/open-telemetry/opentelemetry-go-build-tools/pull/360#discussion_r1258613553 as Option A, these are two separate changelogs, even if on the same file. Sorry, I feel like the original issue was not clear enough as to what each option entailed.

I see what you are saying. I didn't appreciate the difference between Options A and B, except the one file vs two aspect. It seems I am suggesting Option A, though I think having a single file is better.

mx-psi: > What's the case for documenting library changes that are not library API changes? In my mind, a library change that is not a library API change is unremarkable except in the way that it impacts users.

I don't think these are important for opentelemetry-collector-contrib, but I guess I was thinking about other users of the tool (opentelemetry-go) where we do care about library changes. But I guess if we are making something specific to contrib this is not important.

> I see what you are saying. I didn't appreciate the difference between Options A and B, except the one file vs two aspect. It seems I am suggesting Option A is necessary, though I think having a single file is better.

Should we go back to the original issue to decide? I am not sure how to move this forward

mx-psi:
Sounds like the majority opinion on https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/24014 moved towards Option A, so I am closing this and we can continue over at #364

PR #363, opened 2023-07-12
Add the --ignore flag to dbotconf

### LINKED ISSUE #362: Add ignore flag to dbotconf
The https://github.com/open-telemetry/opentelemetry-go-instrumentation repository would like to use `dbotconf`, but there is a `LICENSE` directory that needs to be ignored: https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/104

### Proposal

Add a `--ignore` flag to the root command of `dbotconf` that accepts a comma delimited slice of directory paths that will be ignored during generation and verification.

MrAlias:
Example use:

```shell
dbotconf --ignore "/LICENSES" generate
```

Resolves #362 (verified locally)


internal/repo/repo.go
```diff
@@ -117,6 +128,17 @@ func FindFilePatternDirs(root, pattern string) ([]string, error) {
 			return walkErr
 		}
 
+		for _, p := range ignore {
+			ptrn := filepath.Join(root, p)
```
pellared, line 132: nit: just to have the same as above 😉 

```suggestion
			pattern := filepath.Join(root, p)
```

pellared (approved):

internal/repo/repo.go
MrAlias: This will shadow the function parameter and fail the lint test.

dbotconf/internal/cmd.go
```diff
@@ -43,9 +43,13 @@ var (
 	}
 )
 
+const ignoreFlag = "ignore"
+
 func BuildAndExecute() error {
 	rootCmd.AddCommand(generateCmd)
 	rootCmd.AddCommand(verifyCmd)
 
+	rootCmd.PersistentFlags().StringSlice(ignoreFlag, nil, "directory paths to ignore")
```
mx-psi, line 52: nit: this is a bit ambiguous, it's not really a list of directories but a list of glob patterns as seen here https://github.com/open-telemetry/opentelemetry-go-build-tools/blob/26c6685bcd4658a83564db5fdea67de2c160cb88/internal/repo/repo.go#L74

so we probably want something like:

```suggestion
	rootCmd.PersistentFlags().StringSlice(ignoreFlag, nil, "glob patterns to ignore")
```

or change the implementation to do equality comparison after making paths absolute.

mx-psi (approved):
LGTM, non-blocking comment about the flag docs

dmathieu (approved):

pellared (approved):

PR #374, opened 2023-07-25
semconvgen: Quickfix after semconv repo migration

pellared:
This is "quickfix" for `semconvgen` so that the OTel Collector and Go can use `semconvgen` after the semantic conventions moved from https://github.com/open-telemetry/opentelemetry-specification to https://github.com/open-telemetry/semantic-conventions.

This change was tested as part of https://github.com/open-telemetry/opentelemetry-go/pull/4362

A more sophisticated way would be probably to have two args in  `semconvgen`; one for the Git repository and second for folder when the YAML definitions reside. However, I think that it may be not worth the effort and could introduce more problems than benefits (both OTel Collector and Go would need to update their semconv generation scripts).

### DIFF
--- a/.chloggen/semconvgen-quickfix.yaml
+++ b/.chloggen/semconvgen-quickfix.yaml
@@ -1,11 +1,11 @@
 # One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
-change_type: enhancement
+change_type: bug_fix
 
 # The name of the component, or a single word describing the area of concern, (e.g. crosslink)
 component: semconvgen
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Make semconvgen working for the new semantic conventions repository.
+note: Fix semconvgen to work for the new semantic conventions repository.
 
 # One or more tracking issues related to the change
 issues: [374]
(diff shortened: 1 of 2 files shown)


mx-psi (approved):

.chloggen/semconvgen-quickfix.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. crosslink)
+component: semconvgen
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Make semconvgen working for the new semantic conventions repository.
```
MrAlias, line 8:
```suggestion
note: Fix semconvgen to work for the new semantic conventions repository.
```

.chloggen/semconvgen-quickfix.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
```
MrAlias, line 2:
```suggestion
change_type: bug_fix
```

MrAlias (approved):

PR #379, opened 2023-07-27
[chore] Add checkfile to versions.yaml

pellared:
## Why

`- go.opentelemetry.io/build-tools/checkfile` was missing which is needed during the release process.

## What

1. Update the missing entry in `versions.yaml`.
2. Add `versions.yaml` verification to GH workflow.
3. `Makefile` so that it not happens next time even (safe-guard if e.g. CI does not catch the problem).


MrAlias (approved):

MrAlias:
FYI this will conflict with https://github.com/open-telemetry/opentelemetry-go-build-tools/pull/380

songy23 (approved):

PR #402, opened 2023-09-18
chloggen: add component validation

### LINKED ISSUE #401: Offer to formalize changelog entries by validating the component field
### Component(s)

_No response_

### Is your feature request related to a problem? Please describe.

The current changelog entry format sets a field for the component to which the changes apply. However, this field is free-form and doesn't allow us to formally find if the component is spelled correctly. Often, we use informal names such as "carbonreceiver".

### Describe the solution you'd like

~Ideally, by using class ("receiver", "processor", etc) and type ("carbon", "awss3", ...) we can then create a changelog that is more structured and identifies precisely the component impacted by the change.~

After discussions with maintainers, it appears a simple approach consists of using the go package associated with the component instead, since it is not a collector specific mechanism.

I would like to offer a non-breaking change to the chloggen tool that allows folks to choose to use component just like today, or ~class and type~ go package. When rendering the changelog, ~the class and type fields are used if the component field is left empty, and form a string such as class/type~, the go package is used instead. We can configure chloggen to enforce the use of go package down the road.

The change is offered on this PR: https://github.com/open-telemetry/opentelemetry-go-build-tools/pull/326

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

atoulme:
Fixes #401

### DIFF
--- a/chloggen/cmd/validate_test.go
+++ b/chloggen/cmd/validate_test.go
@@ -142,7 +142,7 @@ func TestValidate(t *testing.T) {
 		{
 			name: "gomodule_validation",
 			cfgFn: func(cfg *config.Config) {
-				cfg.ComponentPrefixes = []string{"github.com/foo/bar/receiver", "github.com/foo/bar/exporter"}
+				cfg.Components = []string{"github.com/foo/bar/receiver", "github.com/foo/bar/exporter"}
 			},
 			entries: func() []*chlog.Entry {
 				sampleEntries := getSampleEntries()
@@ -151,7 +151,7 @@ func TestValidate(t *testing.T) {
 				}
 				return sampleEntries
 			}(),
-			wantErr: "Error: foo is not a valid 'component'. It must start with one of \\[github.com/foo/bar/receiver github.com/foo/bar/exporter\\]",
+			wantErr: "Error: foo is not a valid 'component'. It must be one of \\[github.com/foo/bar/receiver github.com/foo/bar/exporter\\]",
 		},
 	}
 	for _, tc := range tests {
--- a/chloggen/cmd/validate_test.go
+++ b/chloggen/cmd/validate_test.go
@@ -48,15 +48,18 @@ func TestValidateErr(t *testing.T) {
 func TestValidate(t *testing.T) {
 	tests := []struct {
 		name    string
+		cfg     *config.Config
 		entries []*chlog.Entry
 		wantErr string
 	}{
 		{
 			name:    "all_valid",
+			cfg:     config.New(t.TempDir()),
 			entries: getSampleEntries(),
 		},
 		{
 			name: "invalid_change_type",
+			cfg:  config.New(t.TempDir()),
 			entries: func() []*chlog.Entry {
 				return append(getSampleEntries(), &chlog.Entry{
 					ChangeType: "fake",
@@ -69,6 +72,7 @@ func TestValidate(t *testing.T) {
 		},
 		{
 			name: "missing_component",
+			cfg:  config.New(t.TempDir()),
 			entries: func() []*chlog.Entry {
 				return append(getSampleEntries(), &chlog.Entry{
 					ChangeType: chlog.BugFix,
@@ -81,6 +85,7 @@ func TestValidate(t *testing.T) {
 		},
 		{
 			name: "empty_component",
+			cfg:  config.New(t.TempDir()),
 			entries: func() []*chlog.Entry {
 				return append(getSampleEntries(), &chlog.Entry{
 					ChangeType: chlog.BugFix,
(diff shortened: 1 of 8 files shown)


chloggen/internal/config/config.go
```diff
@@ -29,13 +29,16 @@ const (
 	DefaultTemplateYAML      = "TEMPLATE.yaml"
 	DefaultChangeLogKey      = "default"
 	DefaultChangeLogFilename = "CHANGELOG.md"
+
+	DefaultValidationMode = "none"
```
pellared, line 33: Unused?

```suggestion
```

chloggen/internal/chlog/entry.go
```diff
@@ -81,6 +81,18 @@ func (e Entry) Validate(requireChangelog bool, validChangeLogs ...string) error
...
+	found := !(len(componentPrefixes) > 0)
+	for _, prefix := range componentPrefixes {
+		if strings.HasPrefix(e.Component, prefix) {
+			found = true
+			break
+		}
+	}
+	if !found {
+		return fmt.Errorf("%s is not a valid 'component'. It must start with one of %v", e.Component, componentPrefixes)
+	}
```
pellared, line 94: I find the first usage of `found` unreadable.
How about simply

```suggestion
	found := false
	for _, prefix := range componentPrefixes {
		if strings.HasPrefix(e.Component, prefix) {
			found = true
			break
		}
	}
	// only apply prefix validation if one or more prefix is present.
	if len(componentPrefixes) > 0 && !found {
		return fmt.Errorf("%s is not a valid 'component'. It must start with one of %v", e.Component, componentPrefixes)
	}
```

or

```suggestion
	// only apply prefix validation if one or more prefix is present.
	found := len(componentPrefixes) == 0
	for _, prefix := range componentPrefixes {
		if strings.HasPrefix(e.Component, prefix) {
			found = true
			break
		}
	}
	if !found {
		return fmt.Errorf("%s is not a valid 'component'. It must start with one of %v", e.Component, componentPrefixes)
	}
```

chloggen/cmd/validate_test.go
```diff
@@ -48,15 +48,18 @@ func TestValidateErr(t *testing.T) {
 func TestValidate(t *testing.T) {
 	tests := []struct {
 		name    string
+		cfg     *config.Config
```
pellared, line 51: How about changing this to something like
```suggestion
		cfgFn     func(*config.Config)
```

It looks like `config.New(t.TempDir())` is common for all test cases and only some tests needs to have the config modified.

chloggen/internal/config/config.go
```diff
@@ -29,13 +29,16 @@ const (
...
 )
 
 type Config struct {
 	ChangeLogs        map[string]string `yaml:"change_logs"`
 	DefaultChangeLogs []string          `yaml:"default_change_logs"`
 	EntriesDir        string            `yaml:"entries_dir"`
 	TemplateYAML      string            `yaml:"template_yaml"`
+	ComponentPrefixes []string          `yaml:"component_prefixes"`
```
pellared, line 41: Any docs for it? I am guessing that it should be documented here: `chloggen/internal/config/config.yaml`

chloggen/cmd/validate_test.go
atoulme: Fixed, thanks

chloggen/internal/config/config.go
atoulme: Good point, added some

pellared (approved):

djaglowski (commented):
We [previously discussed](https://github.com/open-telemetry/opentelemetry-go-build-tools/issues/401#issuecomment-1722613746) making this validate that the "component" is a valid module. I still think this would be better since it applies more generally.

Perhaps we could just list the actual components which are allowed in the config? (In the future, we could possibly autodetect valid modules, but perhaps it's not for this tool to do that.) In any case, I think listing valid components (ie modules) is much more aligned with the natural way a Go project should organize its release notes.

atoulme:
Yes @djaglowski but I don't know how to check if a string is a valid go module, and the prefix seemed like a first good step to get there. Do you know how to validate this?

chloggen/internal/config/config.yaml
```diff
@@ -21,3 +21,7 @@
 # If 'change_logs' is specified in this file, and no value is specified for 'default_change_logs',
 # then 'change_logs' MUST be specified in every entry file.
 # default_change_logs: []
+
+# The prefixes accepted as valid for the component field of a changelog entry.
+# By default, no prefix is present and any value is valid.
+# component_prefixes: []
```
djaglowski, line 27:
```suggestion
# The valid values for the component field of a changelog entry.
# By default, no value present and any value is valid.
# components: []
```

@atoulme, could we do this instead? It'd be verbose but the changelog would be cleaner. Maybe for repos with many modules we will autogenerate the list.

atoulme: Done!

djaglowski (approved):

pellared (approved):

PR #404, opened 2023-09-22
[multimod] change return code for diff

codeboten:
This matches the behaviour it was intended on replacing in the collector repository. It returns an error if there's no diff to ensure the build does not proceed with the release in that case.

This fixes https://github.com/open-telemetry/opentelemetry-collector/issues/8149

### DIFF
--- a/.chloggen/multimod-diff-rc.yaml
+++ b/.chloggen/multimod-diff-rc.yaml
@@ -1,0 +1,9 @@
+change_type: bugfix
+
+component: multimod
+
+note: Changes the return code for diff to match the behaviour it is replacing
+
+issues: []
+
+subtext:
--- a/.chloggen/multimod-diff-rc.yaml
+++ b/.chloggen/multimod-diff-rc.yaml
@@ -4,6 +4,6 @@ component: multimod
 
 note: Changes the return code for diff to match the behaviour it is replacing
 
-issues: []
+issues: [404]
 
 subtext:
--- a/.chloggen/multimod-diff-rc.yaml
+++ b/.chloggen/multimod-diff-rc.yaml
@@ -1,4 +1,4 @@
-change_type: bugfix
+change_type: bug_fix
 
 component: multimod
 
--- a/.chloggen/multimod-diff-rc.yaml
+++ b/.chloggen/multimod-diff-rc.yaml
@@ -2,8 +2,8 @@ change_type: bug_fix
 
 component: multimod
 
-note: Changes the return code for diff to match the behaviour it is replacing
+note: Return a non-0 exit code if there are no changes to ensure the build does not proceed with the release
 
 issues: [404]
 
-subtext:
+subtext:  This changes the exit code for `diff` to match the behaviour it is replacing in the Collector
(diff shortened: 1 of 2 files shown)


.chloggen/multimod-diff-rc.yaml
```diff
@@ -0,0 +1,9 @@
+change_type: bugfix
+
+component: multimod
+
+note: Changes the return code for diff to match the behaviour it is replacing
+
+issues: []
```
codeboten, line 7:
```suggestion
issues: [404]
```

codeboten:
Pinging @mx-psi for a review :)

.chloggen/multimod-diff-rc.yaml
```diff
@@ -0,0 +1,9 @@
+change_type: bugfix
```
codeboten, line 1:
```suggestion
change_type: bug_fix
```

.chloggen/multimod-diff-rc.yaml
```diff
@@ -0,0 +1,9 @@
+change_type: bug_fix
+
+component: multimod
+
+note: Changes the return code for diff to match the behaviour it is replacing
```
pellared, line 5: This does not tell much for non-Collector repositories 😉

multimod/cmd/diff.go
```diff
@@ -32,9 +32,10 @@ var diffCmd = &cobra.Command{
 		}
 
 		if len(changedFiles) > 0 {
-			log.Fatalf("The following files changed in %s modules since %s: \n%s\nRelease is required for %s modset", moduleSetName, previousVersion, strings.Join(changedFiles, "\n"), moduleSetName)
+			log.Printf("The following files changed in %s modules since %s: \n%s\nRelease is required for %s modset", moduleSetName, previousVersion, strings.Join(changedFiles, "\n"), moduleSetName)
+		} else {
+			log.Fatalf("No %s modules have changed since %s", moduleSetName, previousVersion)
```
pellared, line 37: The exit code is not obvious. [`git diff --exit-code`](https://git-scm.com/docs/git-diff) works the other way. Maybe you could add a description via `cobra.Command.Long` field?

.chloggen/multimod-diff-rc.yaml
mx-psi: What about what you put in the PR description?

```suggestion
note: Return a non-0 exit code if there are no changes to ensure the build does not proceed with the release
```

.chloggen/multimod-diff-rc.yaml
```diff
@@ -0,0 +1,9 @@
...
+
+issues: [404]
+
+subtext:
```
mx-psi, line 9: we can put the current description as subtext.

```suggestion
subtext:  This changes the exit code for `diff` to match the behaviour it is replacing in the Collector
```

.chloggen/multimod-diff-rc.yaml
pellared: Also we should say that 0 exit code is returned when there are changes.

codeboten:
Thanks for the reviews @pellared & @mx-psi. This functionality change always felt a bit strange to me, as the return for `diff` is the opposite of other `diff` tools. I've decided to close the PR and change the logic in the collector's build pipeline to accommodate for the behaviour that's consistent with other tools.

PR #422, opened 2023-10-13
[multimod] add support for --commit-hash to sync

codeboten:
This allows end users to pass in an optional argument when running `multimod sync` if they want to fetch a version that matches a specific tag. This will be useful in the collector contrib for example to sync the core modules, without the user having to manually set the versions in versions.yaml.


.chloggen/codeboten_multimod-add-tag-support.yaml
```diff
@@ -0,0 +1,18 @@
...
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. crosslink)
+component: multimod
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "add support for `--tag` to allow users to overwrite the tag in a versions.yaml file"
+
+# One or more tracking issues related to the change
+issues: []
```
codeboten, line 11:
```suggestion
issues: [422]
```

multimod/internal/sync/sync_test.go
```diff
@@ -177,6 +182,152 @@ func TestNewSync(t *testing.T) {
 	}
 }
 
+type roundTripFunc func(req *http.Request) *http.Response
+
+func (f roundTripFunc) RoundTrip(req *http.Request) (*http.Response, error) {
+	return f(req), nil
+}
+
+func TestUpdateAllGoModFilesWithTag(t *testing.T) {
```
songy23, line 191: Test is failing: https://github.com/open-telemetry/opentelemetry-go-build-tools/actions/runs/6512470406/job/17690170986?pr=422

multimod/internal/sync/sync.go
```diff
@@ -105,12 +110,31 @@ func newSync(myVersioningFilename, otherVersioningFilename, modSetToUpdate, myRe
...
+	body, err := io.ReadAll(res.Body)
+	if err != nil {
+		return "", nil
+	}
+	var data map[string]interface{}
+	err = json.Unmarshal(body, &data)
+	if err != nil {
+		return "", fmt.Errorf("failed to unmarshal response: %w", err)
+	}
+	return fmt.Sprint(data["Version"]), err
```
mx-psi, line 135: We should check if `Version` is in the output instead of panicking, either like this:

```suggestion
	var data map[string]interface{}
	err = json.Unmarshal(body, &data)
	if err != nil {
		return "", fmt.Errorf("failed to unmarshal response: %w", err)
	}
	
	if _, ok := data["Version"]; !ok {
	     return "", fmt.Errorf("response does not contain 'Version' key: %v", data) 
	}
	
	return fmt.Sprint(data["Version"]), err
```

Or maybe something like this (haven't tested the code but should roughly work)

```suggestion
	var data struct{ Version string }
	err = json.Unmarshal(body, &data)
	if err != nil {
		return "", fmt.Errorf("failed to unmarshal response: %w", err)
	}
	
	return fmt.Sprint(data.Version), err
```

multimod/internal/sync/sync.go
```diff
@@ -120,10 +144,20 @@ func (s sync) updateAllGoModFiles() error {
...
 
+	ver := s.OtherModuleSet.Version
+	if s.OtherModuleVersionTag != "" {
+		version, err := s.parseVersionInfo(string(s.OtherModuleSet.Modules[0]), s.OtherModuleVersionTag)
+		if err != nil {
+			return err
+		}
+		ver = version
+	}
+	log.Printf("Version: %s\n", ver)
```
mx-psi, line 155: Stray logging? Should we remove it?

codeboten: This was left here intentionally to make it obvious to the person running the command what version was used in the end

Aneurysm9 (commented):
The use of `tag` here is somewhat confusing to me.  Any change that should have been tagged in a repo using multimod should already have that tag in versions.yaml because that would have been used to create the tags in the first place.  It seems what's intended here is to enable specifying a branch name or commit hash for resolution to a [pseudo-version](https://go.dev/ref/mod#pseudo-versions).  I fully support that, but would like to see the `tag` nomenclature changed to something more reflective of its intent.

multimod/internal/sync/sync.go
```diff
@@ -88,12 +91,14 @@ Then, if necessary, commit changes and push to upstream/make a pull request.`)
 
 // sync holds fields needed to update one module set at a time.
 type sync struct {
-	OtherModuleSetName string
-	OtherModuleSet     common.ModuleSet
-	MyModuleVersioning common.ModuleVersioning
+	OtherModuleSetName    string
+	OtherModuleVersionTag string
```
Aneurysm9, line 95:
```suggestion
	OtherModuleVersionCommit string
```

multimod/internal/sync/sync.go
```diff
@@ -105,12 +110,32 @@ func newSync(myVersioningFilename, otherVersioningFilename, modSetToUpdate, myRe
 	}
 
 	return sync{
-		OtherModuleSetName: modSetToUpdate,
-		OtherModuleSet:     otherModuleSet,
-		MyModuleVersioning: myModVersioning,
+		OtherModuleSetName:    modSetToUpdate,
+		OtherModuleSet:        otherModuleSet,
+		MyModuleVersioning:    myModVersioning,
+		OtherModuleVersionTag: otherVersionCommit,
```
Aneurysm9, line 116:
```suggestion
		OtherModuleVersionCommit: otherVersionCommit,
```

multimod/internal/sync/sync.go
```diff
@@ -120,10 +145,20 @@ func (s sync) updateAllGoModFiles() error {
 		modFilePaths = append(modFilePaths, filePath)
 	}
 
+	ver := s.OtherModuleSet.Version
+	if s.OtherModuleVersionTag != "" {
+		version, err := s.parseVersionInfo(string(s.OtherModuleSet.Modules[0]), s.OtherModuleVersionTag)
```
Aneurysm9, line 150:
```suggestion
	if s.OtherModuleVersionCommit != "" {
		version, err := s.parseVersionInfo(string(s.OtherModuleSet.Modules[0]), s.OtherModuleVersionCommit)
```

multimod/internal/sync/sync_test.go
```diff
@@ -187,6 +338,7 @@ func TestUpdateAllGoModFiles(t *testing.T) {
 	testCases := []struct {
 		modSetName             string
 		expectedOutputModFiles map[string][]byte
+		tag                    string
```
Aneurysm9, line 341:
```suggestion
		commit                    string
```

multimod/internal/sync/sync_test.go
```diff
@@ -336,6 +488,7 @@ func TestUpdateAllGoModFiles(t *testing.T) {
 				otherVersioningFilename,
 				tc.modSetName,
 				tmpRootDir,
+				tc.tag,
```
Aneurysm9, line 491:
```suggestion
				tc.commit,
```

Aneurysm9 (approved):
Some minor nits for remaining references to `tag` instead of `commit`, but otherwise LGTM.

multimod/internal/sync/sync.go
codeboten: thanks for catching, fixed

multimod/internal/sync/sync_test.go
codeboten: fixed

multimod/internal/sync/sync.go
codeboten: fixed

multimod/internal/sync/sync_test.go
codeboten: fixed

multimod/internal/sync/sync_test.go
codeboten: Test is now fixed

mx-psi (approved):

PR #442, opened 2023-11-22
multimod: ignore `excluded-modules` when `-a` flag is enabled

codeboten:
This allows users of the `sync` command to sync all modules in a monorepo, including those listed in the excluded-modules. This is useful for repositories where some modules may not yet be ready for releasing (therefore listed under excluded-modules) but their dependencies still need to be managed via multimod.

### DIFF
--- a/.chloggen/codeboten_add-ignore-excluded.yaml
+++ b/.chloggen/codeboten_add-ignore-excluded.yaml
@@ -5,7 +5,7 @@ change_type: enhancement
 component: multimod
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: add ignore-excluded flag to support 
+note: ignore excluded-modules when -a flag is enabled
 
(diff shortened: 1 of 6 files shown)


atoulme (approved):

pellared (commented):
Do we need this flag? Is there any reason why it shouldn't we simply change the behavior without introducing the flag?

Besides, I find the flag name confusing. Maybe something like `sync-excluded`  or `with-excluded` would be more obvious.

codeboten:
> Do we need this flag? Is there any reason why it shouldn't we simply change the behavior without introducing the flag?

It's a good question, I didn't want to assume that all users would benefit from this change, but thinking back on it, maybe it is desirable for all users to update sync deps even for excluded modules. I could go either ways with this.

codeboten:
@pellared after re-thinking this, I changed the behaviour for anyone that uses the -a flag

pellared (approved):

PR #448, opened 2023-12-05
dbotconf: Fix Windows produces backslashes instead of slashes

### LINKED ISSUE #264: dbotconf: on Windows it produces backslashes instead of slashes
Using `dbotconf` on Windows results in having backslashes in `directory` field instead of slashes

E.g. 

```
  - package-ecosystem: docker
    directory: \example\zipkin
```

instead of

```
  - package-ecosystem: docker
    directory: /example/zipkin
```

I guess the code misses [`filepath.ToSlash()`](https://pkg.go.dev/path/filepath#ToSlash) in some place(s).

scorpionknifes:
Fixes #264

@pellared 
> `multimod tag` is also not working correctly because of backslashes on Windows.
> 
> I guess `crosslink` will have the same issue.

Shall look into `multimod` and `crosslink`


pellared (commented):
First of all thanks for the PR 👍 

1. I see that there are some build failures.
2. What do you think about adding a GitHub Workflow (or job in existing workflow - probably this would be better) that would execute tests on Windows?

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 # Changelog
 
 <!-- next version -->
+- `dbotconf`: Fix Windows produces backslashes instead of slashes (#264)
```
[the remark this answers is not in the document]
pellared: I created https://github.com/open-telemetry/opentelemetry-go-build-tools/issues/449

.gitignore
```diff
@@ -14,10 +14,18 @@ crosslink/internal/test_data/
 !crosslink/internal/test_data/.placeholder
 
 checkfile/checkfile
+checkfile/checkfile.exe
```
pellared, line 17: You can change it to a single rule:

```
*.exe
```

internal/syncerror/known_sync_error_windows.go
```diff
@@ -17,12 +17,16 @@
...
+	"golang.org/x/sys/windows"
+)
 
 // knownSyncError returns true if the given error is one of the known
 // non-actionable errors returned by Sync on Windows:
 //
 // - sync /dev/stderr: The handle is invalid.
 func KnownSyncError(err error) bool {
-	return err == windows.ERROR_INVALID_HANDLE
+	return errors.Is(err, windows.ERROR_INVALID_HANDLE)
```
scorpionknifes, line 31:
```
internal\syncerror\known_sync_error_windows.go:27:9: comparing with == will fail on wrapped errors. Use errors.Is to check for a specific error (errorlint)
        return err == windows.ERROR_INVALID_HANDLE
```
Interesting, didn't know golangci-lint doesn't check go files for Windows when ran on unix systems (ci).

pellared:
> Shall look into multimod and crosslink.

Can you quickly test it and create issues if these tools have the same bug?

pellared (approved):

pellared:
Thanks a lot for your PR 👍 Awaiting second review.

mx-psi (approved):

PR #480, opened 2024-01-15
[crosslink] add --skip flag

codeboten:
This allows modules in monorepos to opt out of crosslink. This will be used in mdatagen in the Collector core repo for example, where mdatagen has dependencies on internal collector modules, but we don't want to use replace statements as it breaks the ability to `go install` mdatagen

Related to https://github.com/open-telemetry/opentelemetry-collector/issues/9281

### DIFF
--- a/crosslink/cmd/root.go
+++ b/crosslink/cmd/root.go
@@ -141,7 +141,7 @@ func init() {
 	comCfg.rootCommand.PersistentFlags().BoolVarP(&comCfg.runConfig.Verbose, "verbose", "v", false, "verbose output")
 	comCfg.rootCommand.Flags().StringSliceVar(&comCfg.excludeFlags, "exclude", []string{}, "list of comma separated go modules that crosslink will ignore in operations."+
 		"multiple calls of --exclude can be made")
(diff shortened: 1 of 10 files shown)


crosslink/cmd/root.go
```diff
@@ -139,6 +141,8 @@ func init() {
 	comCfg.rootCommand.PersistentFlags().BoolVarP(&comCfg.runConfig.Verbose, "verbose", "v", false, "verbose output")
 	comCfg.rootCommand.Flags().StringSliceVar(&comCfg.excludeFlags, "exclude", []string{}, "list of comma separated go modules that crosslink will ignore in operations."+
 		"multiple calls of --exclude can be made")
+	comCfg.rootCommand.Flags().StringSliceVar(&comCfg.skipFlags, "skip", []string{}, "list of comma separated go.mod files that crosslink will ignore in this repo."+
```
pellared, line 144: It looks like `exclude`.

pellared (commented):
I think that `crosslink/README.md` should be updated.

Also for the current description make it hard to understand the difference between `--skip` and `--exclude`.

crosslink/cmd/root.go
codeboten: I've updated the text w/ your suggestion

codeboten:
The readme has been updated, PTAL

TylerHelmuth (approved):

pellared (approved):

Issue #488, opened 2024-02-06
[chloggen] link to issues in changelog


andreasgerstmayr:
It would be handy if the issue number in the changelog entry would be a link to the issue.

PR #489, opened 2024-02-06
[chloggen] Support linking to the issue in the changelog

andreasgerstmayr:
Adds a new configuration setting `issue_link`, which specifies a template where the issues are hosted, for example: `https://github.com/open-telemetry/opentelemetry-go-build-tools/issues/{{issue}}`.

If this setting is specified, the issue numbers in the changelog will be formatted as links to the issue. If the setting is not specified, there is no change how the changelog entries are printed currently.

Resolves: #488


djaglowski (commented):
While I really like the utility this provides, I am concerned it may make it difficult to further generalize the rendering of the changelog.

I've always hoped that we could enable repositories to provide their own template file in place of [`summary.tmpl`](https://github.com/open-telemetry/opentelemetry-go-build-tools/blob/main/chloggen/internal/chlog/summary.tmpl), but never found the time to work on this. It may actually be quite simple but requires a bit of documentation to explain precisely what values are passed to the template.

Anyways, this would allow for exactly what you're looking for but also many other useful customizations. Any chance you're interested in pursuing this approach instead?

djaglowski:
Thanks @andreasgerstmayr!

andreasgerstmayr:
Superseded by #501.

PR #501, opened 2024-02-14
Support custom summary.tmpl file

### LINKED ISSUE #487: [chloggen] Missing line break before subtext
The `subtext` gets rendered in the same paragraph as the `note` (see last two entries):

grafik

Is this expected? I would suggest either adding a line break before the subtext:

* `postgresqlreceiver`: add schema attribute to postgresqlreceiver (#29559)
  Adds a new resource attribute to the PSQL receiver to store the schema of the table or index Existing table attributes are adjusted to not include the schema, which was inconsistently used

Alternatively, if every entry should resemble a paragraph, add a dot after the first sentence:

* `postgresqlreceiver`: add schema attribute to postgresqlreceiver (#29559). Adds a new resource attribute to the PSQL receiver to store the schema of the table or index Existing table attributes are adjusted to not include the schema, which was inconsistently used

### LINKED ISSUE #488: [chloggen] link to issues in changelog

andreasgerstmayr:
Support providing a custom `summary.tmpl` file.

@djaglowski is this what you had in mind for https://github.com/open-telemetry/opentelemetry-go-build-tools/pull/489#pullrequestreview-1866260923?

I didn't update any test input or output, so if everything is covered by unit tests, this PR shouldn't change the formatting of any existing changelog.

I noticed the lines here: https://github.com/open-telemetry/opentelemetry-go-build-tools/blob/ad9317af74bb648efb5b3f0c19b4697f0b468e65/chloggen/internal/chlog/summary.go#L57-L61 don't sort the changelog entries (`sort.Strings` should be used to sort them). I can update this PR to sort the entries by `Component` and `Note` if we want them sorted.

Resolves #488
Resolves #487


chloggen/internal/chlog/summary.go
```diff
@@ -18,57 +18,68 @@ import (
 	"bytes"
 	_ "embed"
 	"fmt"
-	"sort"
+	"path/filepath"
 	"text/template"
+
+	"github.com/Masterminds/sprig/v3"
```
djaglowski, line 24: Do we really need this? I'd rather stick to the standard templating package unless there is a clear need to bring in the dependency and expand the surface area.

djaglowski (commented):
Thanks for putting this together @andreasgerstmayr. This is more or less what I had in mind but I would appreciate it if we can stick to Go's built in templating. We currently have no third party dependencies in this tool and I'm not convinced it's an overall benefit to outsource the surface area.

djaglowski:
> I can update this PR to sort the entries by Component and Note if we want them sorted.

Sorting would be great but let's keep it as a separate PR

chloggen/internal/chlog/summary.go
andreasgerstmayr: I thought it might be useful for custom templates. It's the same library helm uses (https://github.com/helm/helm/blob/0d0f91d1ce277b2c8766cdc4c7aa04dbafbf2503/pkg/engine/funcs.go#L26).

For the standard template we only need an `indent` function. I've reverted the dependency and implemented this function now (it's just a 2 liner), maybe we'll revisit this in the future if there is an actual user demand for more functions.

andreasgerstmayr:
I've dropped the external dependency now, and reverted some of the `summary.tmpl` changes, to keep it as simple as it was before.

chloggen/internal/chlog/summary.go
djaglowski: Thanks @andreasgerstmayr. I'm _necessarily_ opposed to adding it later, just prefer not to add things until the need is clear.

andreasgerstmayr:
> This looks good to me. We will need a changelog entry though :)

:smiley: Done.
Thanks for the review!

djaglowski (approved):

PR #582, opened 2024-07-09
[multimod] Resolve version for commit per module instead of per module set

mx-psi:
Fixes open-telemetry/opentelemetry-collector/issues/9827

### DIFF
--- a/multimod/internal/common/tools.go
+++ b/multimod/internal/common/tools.go
@@ -61,7 +61,7 @@ func GetModuleSet(modSetName, versioningFilename string) (ModuleSet, error) {
 
 // updateGoModVersions updates one go.mod file, given by modFilePath, by updating all modules listed in
 // newModPaths to use the newVersion given.
-func updateGoModVersions(modFilePath ModuleFilePath, newModPaths []ModulePath, newVersion string) error {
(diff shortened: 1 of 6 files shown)


dmathieu (approved):

multimod/internal/common/tools.go
```diff
@@ -61,7 +61,7 @@ func GetModuleSet(modSetName, versioningFilename string) (ModuleSet, error) {
 
 // updateGoModVersions updates one go.mod file, given by modFilePath, by updating all modules listed in
 // newModPaths to use the newVersion given.
```
pellared, line 63: Comment needs to be updated to reflect the change.

pellared (approved):

mx-psi:
I tried this out on open-telemetry/opentelemetry-collector-contrib/pull/33977. It works correctly, although it is a bit slower at first and faster at the tidying stage (likely because the Go proxy is caching the version at an earlier stage). The overall time feels about the same.

multimod/internal/common/tools.go
mx-psi: Fixed, thanks!

codeboten (approved):
Thanks @mx-psi!

PR #612, opened 2024-11-06
fix: set default output of chloggen to stdout

### LINKED ISSUE #611: chloggen prints all output to stdout by default
While doing some debugging, I noticed that chloggen outputs to stdout by default.
That has a few issues in the repos that use this tool, e.g. the opentelemetry-collector repos where the output is redirected to file which doesn't work as it should.

Example:
```
chloggen update --config ./path/to/config --dry 1>stdout.md 2>stderr.md
```

This will result in `stdout.md` being empty and `stderr.md` having the changelog dry run content.

mowies:
Fixes #611

This PR sets the default output for `chloggen` to stdout as it would be set to stderr otherwise which has all kinds of issues that follow in the usage of the tool downstream.

### DIFF
--- a/.chloggen/mowies-fix-chloggen-output.yaml
+++ b/.chloggen/mowies-fix-chloggen-output.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. crosslink)
+component: chloggen
+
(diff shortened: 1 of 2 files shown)


mowies:
markdown-link-check is failing because of a regression on their side, nothing that can be fixed here.

odubajDT (approved):

mx-psi (approved):

dmathieu (approved):

dmathieu:
See https://github.com/open-telemetry/opentelemetry-go-build-tools/pull/613 for the markdown-lint-check issue.

PR #642, opened 2024-12-13
Add `crosslink tidylist` command

jade-guiton-dd:
### Context

While looking into [this issue](https://github.com/open-telemetry/opentelemetry-collector/issues/11795), it turned out that the current `make tidy` target on the `collector-contrib` repo does not properly converge: sometimes updates in one module require updates in its dependent modules, so when running `go mod tidy` in alphabetical order, multiple runs may be necessary to fully propagate changes, potentially as many as the longest dependency chain in the repository.

To solve this issue, I decided to write a script that could run `go mod tidy` on all modules in a repo in an order that guarantees full propagation of all changes. I wrote an [initial prototype in Python](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/36723) to check that the theory worked, but as Collector developers may not have Python installed or know Python well enough to maintain the script, it was suggested that I rewrite it in Go.

### Description

This PR adds a new subcommand to `crosslink`: `crosslink tidy`.

(I decided to add my script as a subcommand because `crosslink` already had most of the code necessary to read the intra-repository module dependency graph and because the code may be useful in other multi-module Go repositories (eg. the core Collector repo). If you believe it does not belong in this tool, I can split the code into its own tool, or just add it in `collector-contrib`.)

### Details

This tool handles circular dependencies, although the schedule it generates will grow quadratically with the size of the largest [strongly-connected component](https://en.wikipedia.org/wiki/Strongly_connected_component) in the dependency graph. For reference, in `collector-contrib` there are 2 SCCs with more than 1 module: one with 2, and one with 3, so I don't believe this would be a problem in practice. Nevertheless, to avoid the unexpected addition of circular dependencies, this tool checks modules that have them against a provided allowlist.

The tool also has a flag to run a naive validation step on the computed solution: it simply checks that all possible dependency chains are subsequences of the schedule, so changes in any upstream module should propagate to any other downstream module. This validation runs in half a second on my machine, but I disabled it by default as it could potentially be quite expensive.

Note that this tool makes the assumption that in the presence of a circular dependency of A on itself, A will not need to be tidied multiple times for the same change (ie. we need to guarantee propagation along paths A → B and A → B → C, but not A → B → C → B). Considering [this worrying description of `go mod tidy` never converging](https://go.dev/src/cmd/go/testdata/script/mod_tidy_convergence.txt), I cannot guarantee that this assumption is always true.


crosslink/internal/tidy.go
```diff
@@ -0,0 +1,204 @@
...
+		var file *os.File
+		file, err = os.Open(rc.AllowCircular)
+		if err != nil {
+			return fmt.Errorf("failed to open cicular dependency allowlist: %w", err)
+		}
+		defer file.Close()
+		scanner := bufio.NewScanner(file)
+		for scanner.Scan() {
+			line := scanner.Text()
+			if line != "" {
```
mx-psi, line 59: Could we also skip lines starting with `#`? That way we can have comments

crosslink/internal/tidy.go
```diff
@@ -0,0 +1,204 @@
...
+
+	rootModule, err := identifyRootModule(rc.RootPath)
+	if err != nil {
+		return fmt.Errorf("failed to identify root module: %w", err)
+	}
+
+	// Read circular dependency allowlist
+
+	var allowCircular []string
+	if rc.AllowCircular != "" {
```
mx-psi, line 49: I am not sure if we validate this, but it would be good to not only skip the modules listed here but to also very that these modules have a circular dependency (to make sure the allowlist is valid)

crosslink/internal/tidy.go
```diff
@@ -0,0 +1,204 @@
...
+
+	graph := make(map[string]*graphNode)
+	var modsAlpha []string
+	err = forGoModFiles(rc, func(filePath string, name string, file *modfile.File) error {
+		if !strings.HasPrefix(name, rootModule) {
+			rc.Logger.Debug("ignoring module outside root module namespace", zap.String("mod_name", name))
+			return nil
+		}
+		modsAlpha = append(modsAlpha, name)
+		modPath, _ := strings.CutSuffix(filePath, "/go.mod")
```
mx-psi, line 78: Why do we need to do this?

crosslink/internal/tidy.go
```diff
@@ -0,0 +1,204 @@
...
+	name    string
+	deps    []string
+	index   int
+	sccRoot int
+	onStack bool
+}
+
+func Tidy(rc RunConfig, outputPath string) error {
+	rc.Logger.Debug("crosslink run config", zap.Any("run_config", rc))
+
```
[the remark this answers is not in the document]
jade-guiton-dd: It was intentional, the other subcommands have a similar log. (It's only shown when the `verbose` flag is passed.)

crosslink/internal/tidy.go
jade-guiton-dd: The `strings.CutSuffix` call? The `filePath` contains the relative path to the `go.mod` file of the module, so we remove `/go.mod` to get the path to the module directory.

mx-psi: Should we check the `found` return argument then and fail if it is `false`?

mx-psi: I feel like it is useful in case the surrounding code changes to be able to more easily debug this (if we ever stop filtering for some reason, we will quickly know that this is failing)

crosslink/cmd/root.go
```diff
@@ -118,6 +119,23 @@ func newCommandConfig() *commandConfig {
 	}
 	c.rootCommand.AddCommand(&c.workCommand)
 
+	c.tidyCommand = cobra.Command{
+		Use:   "tidy output_path",
```
mx-psi, line 123: Thinking about the name a bit more, I think this should not be called `tidy`, since, to the typical Go user, it implies the tool will do the actual tidying. Maybe `generate-tidy-list`? Or any other name that makes the purpose of this subcommand a bit more clear

mx-psi (approved):

crosslink/cmd/root.go
jade-guiton-dd: How about `crosslink tidylist`?

PR #639, opened 2024-12-10
feat: add githubgen tool

mowies:
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/36691

PR #656, opened 2025-01-07
Release v0.16.0

jade-guiton-dd:
### 🚀 New components 🚀

- `githubgen`: Moved githubgen tool here from open-telemetry/opentelemetry-collector-contrib (#639)

### 💡 Enhancements 💡

- `crosslink`: Adds a 'tidy' subcommand to generate 'go mod tidy' schedules (#642)


mx-psi:
@MrAlias I think @jade-guiton-dd is interested in releasing #642, not the `githubgen` changes. I guess we could temporarily revert #639, do the release, then apply #639 again?

mowies:
It should be fine to release even with githubgen. the tool works as is for open-telemetry/opentelemetry-collector-contrib, just not for other use cases. That's what the follow up PR is for.
Either way, the tool is not used anywhere as a dependency yet, so it should be fine as is as well.

jade-guiton-dd:
Yes, I was hesitating to do a release because of the follow up PR. It sounds like it should be okay, but we can delay the release if you feel it necessary. I could use a pseudo-version on contrib in the meantime to start integrating #642.

MrAlias:
I feel we need to revert prior to a release is needed given approvals for #639 were made with the assumption that things would be fixed.

dmathieu:
Couldn't we just remove it from versions.yaml so it's not released rather than a full revert?

jade-guiton-dd:
@dmathieu I can try doing a partial release with only `crosslink`. Has that been done before?

dmathieu:
I believe that would require doing the release manually, which doesn't look like a good idea.

jade-guiton-dd:
I'll try the release again excluding just `githubgen`.

PR #662, opened 2025-01-14
Add `--skip` flag to `crosslink tidylist`

jade-guiton-dd:
Integration of `crosslink tidylist` (cf. #642 for context) in collector-contrib encountered [issues](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/37173), because the output depended on the existence of the `cmd/otelcontribcol` and `cmd/oteltestbedcol`, which are autogenerated and `gitignore`d. I had manually filtered them from the output, but their presence still caused changes to the order of the other modules.

This PR introduces a new `--skip` flag to the `crosslink tidylist` subcommand, similar to the one on the root command, which specifies `go.mod` files that should be entirely ignored by the command (omitted from the output, but also ignored when analyzing the dependency graph). Using this in contrib should fix the non-determinism issue.

I also introduced two new tests to help check the determinism of the tool: one checking that the tool defaults to alphabetical order in the absence of dependency constraints, and one checking that `--skip` works as described above.

### DIFF
--- a/.chloggen/crosslink-tidy-v2.yaml
+++ b/.chloggen/crosslink-tidy-v2.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'enhancement'
+
+# The name of the component, or a single word describing the area of concern, (e.g. crosslink)
+component: crosslink
+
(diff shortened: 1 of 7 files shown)


mx-psi (approved):

songy23 (approved):
LGTM

dmathieu (approved):

PR #672, opened 2025-01-28
Refactor(issuegenerator): Ping codeowners on test failures

ArthurSens:
One of our goals for the collector is to be more proactive when detecting flaky tests on `main,` as mentioned in https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/36761. Since tests must pass before getting merged to `main`, it's safe to assume that the test is flaky if it fails on the `main` branch.

Thanks to https://github.com/open-telemetry/opentelemetry-collector/pull/11963, we can easily parse test results and correlate them with the corresponding code owners using the artifacts generated in JUnit format.

---

This PR refactors the `issuegenerator` tool to make use of the mentioned JUnit files and to also ping codeowners when it detects failed tests.


ArthurSens:
I'm opening it as a draft first so people can chime in on the approach :) 

This is still not working 100% because I'm failing to correlate module names with the appropriate filepaths that are present in the codeowners file.

A module name has the full name that is present in the go.mod file, while the pattern used in the codeowners files uses only the relative path of files. 😕

mx-psi (commented):
This looks reasonable to me, have you done any tests with real contrib data? I would be willing to try directly on contrib, if we can restrict it to a just a few modules to begin with (e.g. we could begin with just extensions)

ArthurSens:
I haven't tried with contrib, but with my fork of collector-core. I've just added one [commit updating the codeowners](https://github.com/open-telemetry/opentelemetry-collector/compare/main...ArthurSens:opentelemetry-collector:main) (This test was done before removing the codeowners bit from this PR).

The issue was opened without problems: https://github.com/ArthurSens/opentelemetry-collector/issues/1

> if we can restrict it to a just a few modules to begin with (e.g. we could begin with just extensions)

I can see two ways forward here:
* Figure out a way to only generate JUnit files for the modules we want
* Add another flag here, e.g. a regex matcher and only generate reports for modules that match this regex

mx-psi:
> Figure out a way to only generate JUnit files for the modules we want

I think we can do that using some conditionals on Github Actions, no need to modify this tool

issuegenerator/main.go
```diff
@@ -30,82 +32,144 @@ import (
...
+			if file.IsDir() {
+				continue
+			}
+			rg.logger.Info("Ingesting test reports", zap.String("path", file.Name()))
+			suites, err := junit.IngestFile(path.Join(pathToArtifacts, file.Name()))
+			if err != nil {
+				rg.logger.Warn(
+					"Failed to ingest JUnit xml, omitting test results from report",
+					zap.Error(err),
+				)
```
mx-psi, line 122: I would just fail here, we won't notice the warning otherwise

mx-psi (approved):

issuegenerator/go.mod
```diff
@@ -5,11 +5,22 @@ go 1.22.0
...
 
 require (
+	github.com/kr/pretty v0.3.1 // indirect
+	gopkg.in/check.v1 v1.0.0-20201130134442-10cb98267c6c // indirect
+)
+
+require (
+	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/google/go-querystring v1.0.0 // indirect
+	github.com/hmarr/codeowners v1.2.1
```
ArthurSens, line 21: wait, this should not be here anymore 🤔

issuegenerator/main.go
```diff
@@ -30,82 +32,144 @@ import (
...
-		createdIssueComment := rg.commentOnIssue(existingIssue)
-		rg.logger.Info("GitHub Issue updated", zap.String("html_url", *createdIssueComment.HTMLURL))
+
+		existingIssue := rg.getExistingIssue(report.module)
+		if existingIssue == nil {
+			// If none exists, create a new GitHub Issue for the failure.
+			rg.logger.Info("No existing Issues found, creating a new one.")
+			createdIssue := rg.createIssue(report)
+			rg.logger.Info("New GitHub Issue created", zap.String("html_url", *createdIssue.HTMLURL))
+			rg.reportIterator++
```
mowies, line 72: this can be moved out of the if block

issuegenerator/main.go
```diff
@@ -115,9 +179,9 @@ type reportGenerator struct {
 func (rg *reportGenerator) getRequiredEnv() {
```
mowies, line 179: i think the doc string for this method needs an update

mowies (commented):
- please either use methods that always have the value receiver or the object receiver :) right now, it's a mix of both
- i like it more when types/consts/vars etc. are at the top of the file and then the funcs follow that

issuegenerator/main.go
ArthurSens: Great catch, this also made me research the default env variables from Github actions and realize we can re-use some of them :)

I'll add a commit updating the variables and doc string

ArthurSens:
> * please either use methods that always have the value receiver or the object receiver :) right now, it's a mix of both

Could you elaborate on that? I'm not sure if I understood 😬 

> * i like it more when types/consts/vars etc. are at the top of the file and then the funcs follow that

Last commit should do it, how do you like it?

mowies:
> > * please either use methods that always have the value receiver or the object receiver :) right now, it's a mix of both
> 
> Could you elaborate on that? I'm not sure if I understood 😬

in the current code there's both those cases:
```go
func (rg *reportGenerator) something(foo string)
func (rg reportGenerator) something(foo string)
```
as you can see, one has a pointer there and one doesn't. that should be consistent throughout the code :)

ArthurSens:
Ah of course, thank you!

mowies (approved):
LGTM

andrzej-stencel (approved):

Issue #673, opened 2025-01-30
[githubgen] tool should fill existing codeowners file and fill sections delimited by magic strings


mowies:
Right now, githubgen uses a constant string as the basis to generate the codeowners file. That's not very generic and a better solution would be to use what is already done with issue templates:

The file already exists and has some magic strings inside that are used as delimiters for the sections that githubgen should fill. This approach has the advantage that users can have a custom file that can include custom settings and then githubgen fills in the auto-generated content for components.
An example base codeowners file could look like this:

```
#####################################################
#
# List of codeowners
#
#####################################################
#
# Learn about CODEOWNERS file format:
# https://help.github.com/en/articles/about-code-owners
#

* @/default-owners           # <---- some default section with custom entries that are part of the base file
some/folder @/some-custom-owners

# Start components list      # <---- magic string to start section

...

# End components list        # <---- magic string to end section

#####################################################
#
# List of distribution maintainers
#
#####################################################

# Start distribution list      # <---- magic string to start section

...

# End distribution list        # <---- magic string to end section

#####################################################
#
## UNMAINTAINED components
#
#####################################################

# Start unmaintained components list      # <---- magic string to start section

...

# End unmaintained components list        # <---- magic string to end section
```

PR #674, opened 2025-01-30
[githubgen] inject codeowner list into existing files

mowies:
This PR adds a breaking change to githubgen!
githubgen now injects its content into the CODEOWNERS and ALLOWLIST files instead of writing complete files every time.
For this, magic strings are used to start and end sections that should be filled by githubgen. For example, `# Start component list` and `# End component list`.

This is a complete list with magic strings and the content that is filled in:
- CODEOWNERS file
  - `# Start/End components list`: Components and their codeowners
  - `# Start/End distribution list`: Distributions and their codeowners
  - `# Start/End unmaintained components list`: Unmaintained components and the configured default codeowner
- ALLOWLIST file
  - `# Start/End deprecated components list`: Deprecated components
  - `# Start/End unmaintained components list`: Unmaintained components

Fixes https://github.com/open-telemetry/opentelemetry-go-build-tools/issues/673


atoulme (approved):

mx-psi (approved):

githubgen/codeowners.go
```diff
@@ -208,3 +239,19 @@ func GetGithubMembers(skipGithub bool, githubOrg string) (map[string]struct{}, e
...
+	}
+	return fileContents, nil
+}
+
+func setFile(fileName string, data []byte) error {
+	err := os.WriteFile(fileName, data, 0o600) // nolint: gosec
+	if err != nil {
+		return err
+	}
+	return nil
```
odubajDT, line 256:
```suggestion
	return os.WriteFile(fileName, data, 0o600) // nolint: gosec
```

githubgen/codeowners.go
```diff
@@ -208,3 +239,19 @@ func GetGithubMembers(skipGithub bool, githubOrg string) (map[string]struct{}, e
 	}
 	return usernames, nil
 }
+
+func getFile(fileName string) ([]byte, error) {
+	fileContents, err := os.ReadFile(fileName) // nolint: gosec
+	if err != nil {
+		return nil, err
+	}
+	return fileContents, nil
```
odubajDT, line 248:
```suggestion
	return os.ReadFile(fileName) // nolint: gosec
```

githubgen/codeowners.go
```diff
@@ -32,73 +36,100 @@ func (cg *codeownersGenerator) Generate(data datatype.GithubData) error {
...
+
+	err = cg.setFile(allowListFile, allowListContents)
 	if err != nil {
 		return err
 	}
 	return nil
 }
 
+func injectContent(startMagicString, endMagicString string, templateContents []byte, replaceContent []string) []byte {
+	matchOldContent := regexp.MustCompile("(?s)" + startMagicString + ".*" + endMagicString)
```
odubajDT, line 117: are we sure it will always compile? I would rather use `regexp.Compile` and check for the error in case the regex string is invalid

githubgen/main.go
```diff
@@ -169,7 +169,7 @@ func newIssueTemplatesGenerator(trimSuffixes string) *issueTemplatesGenerator {
 }
 
 func newCodeownersGenerator(skipGithubCheck *bool) *codeownersGenerator {
-	return &codeownersGenerator{skipGithub: *skipGithubCheck, getGitHubMembers: GetGithubMembers}
+	return &codeownersGenerator{skipGithub: *skipGithubCheck, getGitHubMembers: getGithubMembers, getFile: getFile, setFile: setFile}
```
odubajDT, line 172: Maybe you can introduce a NewCodeownersGenerator() func which sets the `setFile` and `getFile` by default (maybe also the other parameters, if they are not from user input)

mowies: doesn't my factory function do exactly that?

odubajDT: oh sorry my bad, read the code wrongly, all good here

githubgen/codeowners.go
mowies: i think it is going to always compile. the regular expression is made up of all constant parts so there should be no variation at all.

odubajDT (approved):

PR #702, opened 2025-02-27
Update dependencies for excluded modules

XSAM:
Part of https://github.com/open-telemetry/opentelemetry-go-contrib/issues/6440

Currently, the `prerelease` command won't update the version of dependencies for excluded modules. This brings an issue where `prerelease` cannot update dependencies for those example packages, which usually won't be released (thus excluded).

This PR fixes this issue by scanning all go modules under the root path regardless of whether they are excluded or not.

Blocked by #703


MrAlias (approved):

pellared:
@open-telemetry/collector-approvers, can you take a look and make sure it also won't cause any problems for the Collector releases?

mx-psi (approved):

PR #738, opened 2025-04-07
[checkapi] support diffs

atoulme:
Add ability to check if the Golang API of a module has changed

### DIFF
--- a/.chloggen/apidiff.yaml
+++ b/.chloggen/apidiff.yaml
@@ -1,0 +1,18 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. crosslink)
+component: checkapi
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add ability to check if the Golang API of a module has changed
+
+# One or more tracking issues related to the change
+issues: [738]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  This adds a new set of rules to offer a diff when an API change is made.
+  This can error out on removing or adding new elements to the API.
--- a/checkapi/internal/diff_test.go
+++ b/checkapi/internal/diff_test.go
@@ -36,7 +36,7 @@ func TestDiff(t *testing.T) {
 		Params:      []string{"string", "bool"},
 	}}, bBeforeFn.Functions...)
 	bAddStruct := b
-	bAddStruct.Structs = append(bAddStruct.Structs, Apistruct{Name: "MyStruct", Fields: []string{"foo", "bar"}})
+	bAddStruct.Structs = append(bAddStruct.Structs, APIstruct{Name: "MyStruct", Fields: []string{"foo", "bar"}})
 
 	tests := []struct {
 		name         string
@@ -112,7 +112,7 @@ func TestDiff(t *testing.T) {
 				Left:  API{},
 				Equal: a,
 				Right: API{
-					Structs: []Apistruct{
+					Structs: []APIstruct{
 						{
 							Name:   "MyStruct",
 							Fields: []string{"foo", "bar"},
@@ -139,7 +139,7 @@ func TestDiff(t *testing.T) {
 				},
 				Equal: a,
 				Right: API{
-					Structs: []Apistruct{
+					Structs: []APIstruct{
 						{
 							Name:   "MyStruct",
 							Fields: []string{"foo", "bar"},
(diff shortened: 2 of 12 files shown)


mx-psi (commented):
Could we do an example PR where we see this output?

.chloggen/apidiff.yaml
```diff
@@ -0,0 +1,18 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. crosslink)
+component: checkapi
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add ability to check if the Golang API of a module has changed
```
pellared, line 8:
```suggestion
note: Add ability to check API compatibility of a Go module
```

pellared (commented):
What about type parameters (generics)?

atoulme:
> What about type parameters (generics)?

I can add a test using generics, that's not come up yet but should be fun to support.

atoulme:
@pellared generics peeled off in their own PR here: https://github.com/open-telemetry/opentelemetry-go-build-tools/pull/739

atoulme:
Peeling off returning by value in https://github.com/open-telemetry/opentelemetry-go-build-tools/pull/741

checkapi/internal/diff_test.go
```diff
@@ -0,0 +1,181 @@
...
+	bBeforeFn.Functions = append([]Function{{
+		Name:        "Aaaa",
+		Receiver:    "foo",
+		ReturnTypes: []string{"string", "error"},
+		Params:      []string{"string", "bool"},
+	}}, bBeforeFn.Functions...)
+	bAddStruct := b
+	bAddStruct.Structs = append(bAddStruct.Structs, Apistruct{Name: "MyStruct", Fields: []string{"foo", "bar"}})
+
+	tests := []struct {
```
pellared, line 41: Consider adding the following test cases:

- adding another  type parameter to an existing function (this is a backwards compatible change)
- adding a new parameter to a function (this is a breaking change)

checkapi/internal/diff_test.go
```diff
@@ -0,0 +1,181 @@
...
+	})
+	bBeforeFn := b
+	bBeforeFn.Functions = append([]Function{{
+		Name:        "Aaaa",
+		Receiver:    "foo",
+		ReturnTypes: []string{"string", "error"},
+		Params:      []string{"string", "bool"},
+	}}, bBeforeFn.Functions...)
+	bAddStruct := b
+	bAddStruct.Structs = append(bAddStruct.Structs, Apistruct{Name: "MyStruct", Fields: []string{"foo", "bar"}})
```
pellared, line 39: It looks like currently you are unable to detect changes related to field types and type parameters, right?

atoulme: you can but it's a bit blunt. I am adding a test for the addition of a field.

If you add a `foobar` field, the result is:

```
				Left: API{
					Structs: []Apistruct{{Name: "MyStruct", Fields: []string{"foo", "bar"}}},
				},
				Equal: API{},
				Right: API{
					Structs: []Apistruct{{Name: "MyStruct", Fields: []string{"foo", "bar", "foobar"}}},
				},
```

checkapi/internal/diff_test.go
atoulme: I added both, but they probably don't do what you want. Right now, the algorithm just finds the api to be not equal. I think you are asking for a different treatment - where you'd get right breaking change, right, equal, left, left breaking change.

pellared:
@atoulme, there are some lint failures. Besides, I guess this should be reviewed by OTel Collector approvers. Have you considered using https://pkg.go.dev/golang.org/x/exp/cmd/gorelease?

atoulme:
> Have you considered using https://pkg.go.dev/golang.org/x/exp/cmd/gorelease?
I think we looked at it and it didn't really meet the requirements.

atoulme:
@open-telemetry/collector-approvers and @open-telemetry/collector-contrib-approvers please review

mx-psi (commented):
I think this functionality is welcome but I would like to understand why we want to have a custom solution instead of relying in https://pkg.go.dev/golang.org/x/exp/apidiff as Kubernetes does

atoulme:
I didn't look at that tool. I don't have an opinion on it. Should we prefer it?

pellared:
> I didn't look at that tool. I don't have an opinion on it. Should we prefer it?

This is something which the Go team uses as well. https://pkg.go.dev/golang.org/x/exp/cmd/gorelease uses this package as well. There are still some gotchas (e.g. see https://github.com/golang/go/issues/26420#issuecomment-2669406841), but I think we should prefer it.

mx-psi:
> I didn't look at that tool. I don't have an opinion on it. Should we prefer it?

I thought it was unmaintained, but talking at KubeCon with someone from Kubernetes they mentioned they use this (see https://github.com/kubernetes/kubernetes/blob/master/hack/apidiff.sh and https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/129125/pull-kubernetes-apidiff-client-go/1887847435644964864 as a concrete example), and that the Go team has been responsive in adding features like support for generics. So I think it's less maintenance work if we leverage it (possibly wrapping it within checkapi)

pellared:
I would also consider contributing to https://pkg.go.dev/golang.org/x/exp/cmd/gorelease. AFAIK the main issue is lack of support for multiple Go modules.

atoulme:
OK, I will close this PR and maybe we can discuss this from the need perspective a bit more in the collector repository, and try any of the two tools you mention. The branch is still there if anyone wants to revisit.

PR #739, opened 2025-04-09
[chore] support generics

atoulme:
This adds generics support when outputting function signatures.

### DIFF
--- a/checkapi/internal/ast_test.go
+++ b/checkapi/internal/ast_test.go
@@ -58,23 +58,24 @@ func TestAST(t *testing.T) {
 		},
 		{
 			"func with params and return type",
-			&ast.FuncType{Params: &ast.FieldList{
-				List: []*ast.Field{
+			&ast.FuncType{
+				Params: &ast.FieldList{
+					List: []*ast.Field{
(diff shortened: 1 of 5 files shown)


checkapi/internal/ast_test.go
```diff
@@ -126,6 +127,30 @@ func TestAST(t *testing.T) {
...
+						Names: []*ast.Ident{
+							ast.NewIdent("foo"),
+						},
+						Type: ast.NewIdent("T"),
+					},
+				},
+				},
+				Results: nil,
+			},
+			"func[T ~string](T) ",
```
MovieStoreGuy, line 152: Is there meant to be a space on the end of the string?

MovieStoreGuy (approved):

checkapi/internal/ast_test.go
atoulme: I probably could try harder and remove that

atoulme: https://github.com/open-telemetry/opentelemetry-go-build-tools/pull/740 is open to follow up.

PR #862, opened 2025-04-29
[issuegenerator] Context should not be stored in struct

crobert-1:
golang [documentation](https://pkg.go.dev/context) states:
```
Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it.
```
This change removes the stored context in favor of passing it between functions.

### DIFF
--- a/issuegenerator/main.go
+++ b/issuegenerator/main.go
@@ -60,7 +60,6 @@ ${failedTests}
 )
 
 type reportGenerator struct {
-	ctx          context.Context
 	logger       *zap.Logger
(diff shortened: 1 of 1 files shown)


crobert-1:
_I don't believe this was causing any bugs at this time, but let me know if we want a changelog anyways._

MrAlias (approved):

pellared (approved):

mx-psi (approved):

andrzej-stencel (approved):

Issue #861, opened 2025-04-29
[issuegenerator] Duplicating issues instead of adding frequency


crobert-1:
In contrib, an [open issue](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/39632) exists for failing integration tests. However, duplicate new issues were recently opened instead of adding frequency to the existing issue.

[Example 1](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/39743)
[Example 2](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/39744)

There's no difference between the new issues and the existing issue.

crobert-1:
It looks like when checking for an existing issue the prefix is being trimmed from the module. When creating the issue, the prefix is left.

Checking for existing issue logic:
https://github.com/open-telemetry/opentelemetry-go-build-tools/blob/8934a537fd7890da8bd9aa3c17f16cda0699b46d/issuegenerator/main.go#L255-L256

Creating a new issue:
https://github.com/open-telemetry/opentelemetry-go-build-tools/blob/8934a537fd7890da8bd9aa3c17f16cda0699b46d/issuegenerator/main.go#L294

This functionality should be consolidated to one or the other. The prefix should be removed in both, or kept in both.

CC: @ArthurSens

PR #864, opened 2025-04-29
[issuegenerator] Trim module name when creating a new issue

crobert-1:
When checking to see if an existing issue is already open for a failed test module, the module name is trimmed to remove the github repo prefix of `github.com/githubOwner/githubRepository/`. However, when creating an issue, the entire module name is being used. Names need to be consistent to ensure duplicates are not filed, and frequency is added to existing issues when necessary. This fixes them to be consistent.

Resolves https://github.com/open-telemetry/opentelemetry-go-build-tools/issues/861

### DIFF
--- a/issuegenerator/main.go
+++ b/issuegenerator/main.go
@@ -94,11 +94,7 @@ func newReportGenerator() *reportGenerator {
 }
 
 func trimModule(owner, repo, module string) string {
-	httpsPrefix := "https://"
-	if strings.HasPrefix(module, httpsPrefix) {
-		module = strings.TrimPrefix(module, httpsPrefix)
-	}
-
+	module = strings.TrimPrefix(module, "https://")
 	return strings.TrimPrefix(module, fmt.Sprintf("github.com/%s/%s/", owner, repo))
 }
 
--- a/issuegenerator/main.go
+++ b/issuegenerator/main.go
@@ -94,7 +94,6 @@ func newReportGenerator() *reportGenerator {
 }
 
 func trimModule(owner, repo, module string) string {
-	module = strings.TrimPrefix(module, "https://")
 	return strings.TrimPrefix(module, fmt.Sprintf("github.com/%s/%s/", owner, repo))
 }
 
--- a/issuegenerator/main.go
+++ b/issuegenerator/main.go
@@ -294,7 +294,8 @@ func (rg *reportGenerator) commentOnIssue(issue *github.Issue) *github.IssueComm
 
 // createIssue creates a new GitHub Issue corresponding to a build failure.
 func (rg *reportGenerator) createIssue(r report) *github.Issue {
-	title := strings.Replace(issueTitleTemplate, "${module}", r.module, 1)
+	trimmedModule := strings.TrimPrefix(r.module, fmt.Sprintf("github.com/%s/%s/", rg.envVariables["githubOwner"], rg.envVariables["githubRepository"]))
+	title := strings.Replace(issueTitleTemplate, "${module}", trimmedModule, 1)
 	body := os.Expand(issueBodyTemplate, rg.templateHelper)
 
 	issue, response, err := rg.client.Issues.Create(
--- a/issuegenerator/main.go
+++ b/issuegenerator/main.go
@@ -93,6 +93,15 @@ func newReportGenerator() *reportGenerator {
 	}
 }
 
+func trimModule(owner, repo, module string) string {
+	https_prefix := "https://"
+	if strings.HasPrefix(module, https_prefix) {
+		module = strings.TrimPrefix(module, https_prefix)
+	}
+
+	return strings.TrimPrefix(module, fmt.Sprintf("github.com/%s/%s/", owner, repo))
+}
+
 func main() {
 	pathToArtifacts := flag.String("path", "", "Path to the directory with test results")
 	flag.Parse()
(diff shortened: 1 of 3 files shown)


issuegenerator/main.go
```diff
@@ -291,7 +291,8 @@ func (rg *reportGenerator) commentOnIssue(issue *github.Issue) *github.IssueComm
 
 // createIssue creates a new GitHub Issue corresponding to a build failure.
 func (rg *reportGenerator) createIssue(r report) *github.Issue {
-	title := strings.Replace(issueTitleTemplate, "${module}", r.module, 1)
+	trimmedModule := strings.TrimPrefix(r.module, fmt.Sprintf("github.com/%s/%s/", rg.envVariables["githubOwner"], rg.envVariables["githubRepository"]))
```
crobert-1, line 294: This was copied from https://github.com/open-telemetry/opentelemetry-go-build-tools/blob/8934a537fd7890da8bd9aa3c17f16cda0699b46d/issuegenerator/main.go#L255. We could alternatively pull out the trimming logic to its own method and pass in the module, github owner, and github repository.

ArthurSens: With https://github.com/open-telemetry/opentelemetry-go-build-tools/pull/863, we should also use constants here.

Regarding moving the logic to a different function, I'm not strongly opinionated here... Whatever feels best for you :)

crobert-1: I've moved to a central method, added testing, and also duplicated the central variables. Once this is rebased on `main` that includes the constants from #863 it will properly show that this PR is not adding the constants, simply adding references to them.

crobert-1:
This requires #863 to be merged first, as it relies on constants defined there.

crobert-1:
I'll mark as ready for review once #863 is merged

issuegenerator/main_test.go
```diff
@@ -108,3 +108,42 @@ func TestIngestArtifacts(t *testing.T) {
 	require.Equal(t, expectedTestResults, rg.testSuites)
 
 }
+
+func TestTrimPath(t *testing.T) {
+
```
pellared, line 113:
```suggestion
```

issuegenerator/main_test.go
```diff
@@ -108,3 +108,42 @@ func TestIngestArtifacts(t *testing.T) {
...
+	tests := []struct {
+		name           string
+		owner          string
+		repo           string
+		module         string
+		expectedModule string
```
pellared, line 119: in Go we usually name the expected value as `want`; reference: https://go.dev/wiki/CodeReviewComments#useful-test-failures

```suggestion
		wantModule string
```

issuegenerator/main_test.go
```diff
@@ -108,3 +108,42 @@ func TestIngestArtifacts(t *testing.T) {
...
+			owner:          "open-telemetry",
+			repo:           "opentelemetry-collector",
+			module:         "github.com/open-telemetry/opentelemetry-collector/exporter/otlphttpexporter/otlp_test.go",
+			expectedModule: "exporter/otlphttpexporter/otlp_test.go",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			require.Equal(t, tt.expectedModule, trimModule(tt.owner, tt.repo, tt.module))
```
pellared, line 146: this would scale better if you have more assertions

```suggestion
			assert.Equal(t, tt.expectedModule, trimModule(tt.owner, tt.repo, tt.module))
```

pellared, line 146: It would be nice to also provide the test case input (owner, repo, module) so when the test fails the reader could see more than just the test name and actual vs expected value. References: 
- https://go.dev/wiki/CodeReviewComments#useful-test-failures
- https://go.dev/wiki/TestComments#identify-the-input

issuegenerator/main_test.go
```diff
@@ -108,3 +108,42 @@ func TestIngestArtifacts(t *testing.T) {
...
+		repo           string
+		module         string
+		expectedModule string
+	}{
+		{
+			name:           "Test contrib host metrics integration path with https prefix",
+			owner:          "open-telemetry",
+			repo:           "opentelemetry-collector-contrib",
+			module:         "https://github.com/open-telemetry/opentelemetry-collector-contrib/receiver/hostmetricsreceiver/integration_test.go",
+			expectedModule: "receiver/hostmetricsreceiver/integration_test.go",
```
ArthurSens, line 126: The module that we get from automation is actually the [module name as written in the go.mod file](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/df65f76b84de0683b892ee5f32e72f8781d44f0a/receiver/hostmetricsreceiver/go.mod#L1), not the test file name. 

But not really important, the trimming is what's being tested and it works 👍

ArthurSens (commented):
It's a LGTM from me :)

We just have some nits here and there

issuegenerator/main_test.go
crobert-1: Thanks for the information. I've removed the prefix check and testing, we can keep it simple for now and change as needed.

issuegenerator/main_test.go
crobert-1: Thanks for the links. I've historically followed the `testify` language of `expected`, but I can see the value in following best practices over a package we're dependent upon. Fixed.

ArthurSens (approved):
Thanks!

PR #894, opened 2025-05-15
[checkapi] check the API of a component

atoulme:
Allow to check the API of components matches strictly to guidelines, with the only structs exported the Config structs.


mx-psi:
```
verifyAllModulesInSet failed: Module github.com/open-telemetry/opentelemetry-go-build-tools/checkapi/internal/config/receiver/configreceiver (defined in /home/runner/work/opentelemetry-go-build-tools/opentelemetry-go-build-tools/checkapi/internal/config/receiver/configreceiver/go.mod) is not listed in any module set.
```

atoulme:
Still needs some love. Iterating against contrib to catch all the cases.

atoulme:
I think we could make an additional flag to ignore errors coming from being able to parse the component, at least temporarily. This way we can get this in and do some good with the obvious issues reported above and avoid scope creep and open an issue to continue the work on making the tool better.

checkapi/internal/ast.go
```diff
@@ -148,9 +180,6 @@ func readFile(ignoredFunctions []string, f *ast.File, result *API) {
 			}
 		}
 		if fn, isFn := d.(*ast.FuncDecl); isFn {
-			if !fn.Name.IsExported() {
```
MovieStoreGuy, line 151: Any reason why this was removed?

MovieStoreGuy (approved):

checkapi/internal/ast.go
atoulme: we only captured functions that were exported, but with this change we need to capture unexported functions such as `createDefaultConfig` so we can source the Config struct.

PR #993, opened 2025-06-30
Default to pre-releasing all module sets in `multimod prerelease`

### LINKED ISSUE #992: Default to all modules in `multimod prerelease`
When releasing, I commonly want most of the module sets to be updated. Currently, this means I need to parse the `versions.yaml` file to determine what needs to be updated. This seems like something the tooling can accomplish for me instead.

Ideally, the `multimod prerelease` would just try to update all module sets by default. Any module set that hasn't been bumped is a no-op, so this behavior would match my expectation.

MrAlias:
Resolve #992 

Attempt to make a pre-release for all module sets so a user doesn't have to go parse the `versions.yaml` file.

If no updates are made for a module set nothing is done so this becomes a no-op for module sets that do not need to be pre-released:

```terminal
Using versioning file /home/tyler/go/src/go.opentelemetry.io/build-tools/versions.yaml
Using repo with root at /home/tyler/go/src/go.opentelemetry.io/build-tools

===== Module Set: tools =====
Module set already up to date (git tags already exist). Skipping...
=========
Prerelease finished successfully. Now checkout the new branch(es) and verify the changes.

Then, if necessary, commit changes and push to upstream/make a pull request.
```

The `--module-set-names` flag can still be used to provide an allow-list of modules if a targeted group needs to be addressed.


.chloggen/prerelease-all.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
```
mx-psi, line 2: Isn't this a breaking change?

mx-psi (approved):
LGTM, the test coverage is lacking but that is not this PR's fault

PR #1052, opened 2025-07-22
[multimod] Use errgroup to parallelize sync requests

mx-psi:
Use [errgroup](https://pkg.go.dev/golang.org/x/sync/errgroup) to parallelize requests to the Go proxy

### DIFF
--- a/multimod/internal/sync/sync.go
+++ b/multimod/internal/sync/sync.go
@@ -21,6 +21,7 @@ import (
 	"io"
 	"log"
 	"net/http"
+	"time"
 
 	"github.com/go-git/go-git/v5"
 	"golang.org/x/sync/errgroup"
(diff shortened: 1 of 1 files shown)


andrzej-stencel (approved):

andrzej-stencel (approved):
Looks good. I think the retryablehttp library does not apply any jitter to the retries by default?

EDIT: My thinking is 1) with jitter, we could end up with total less than 31s and 2) would jitter be actually useful here?

mx-psi:
> Looks good. I think the retryablehttp library does not apply any jitter to the retries by default?

It doesn't look like the default policy has any jitter, no

https://github.com/hashicorp/go-retryablehttp/blob/e1f5485fe84728709b857cb89e17088894c301d6/client.go#L560

> EDIT: My thinking is to consider 1) with jitter, we could end up with total less than 31s and 2) would jitter be actually useful here?

I don't think jitter is useful here, no, and that's good point about jitter messing up my calculations

PR #1064, opened 2025-07-25
issugenerator: Don't fatal when failing to ingest a JUnit file

ArthurSens:
It's better to complete the issue generation with a few reports missing than not do it at all 😬


dmathieu:
To view this warning, one has to open the CI output, which is unlikely to happen if it didn't fail.
If we mention it in the issue comment, it's much more visible.

dmathieu:
Doing this in a follow up is fine.
Just saying the CI output had warnings would be fine too IMHO.

dmathieu (approved):

PR #1104, opened 2025-08-11
Release 0.27.0

atoulme:


mx-psi (approved):

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@
 
 <!-- next version -->
 
+## 0.27.0
```
pellared, line 7: Why is it not 0.26.1 if it is only a bugfix release?

adrielp (approved):
Approved, same comment as @pellared though. Looks like the only change here should be a patch release, not a minor release.

PR #1132, opened 2025-08-21
[chloggen] Accept entries with `.yml` extension

### LINKED ISSUE #1079: Chloggen only picks up .yaml files and not .yml
### Component

chloggen

### Bug Description

In the collector v0.131.0 release, chloggen seemed to not pick up a changelog entry ending with `.yml` instead of `.yaml`. I'm not sure if this is an actual bug or a misconfiguration.

### Steps To Reproduce

- fetch chloggen config from [collector-releases repo](https://github.com/open-telemetry/opentelemetry-collector-releases/blob/main/.chloggen/config.yaml)
- create a changelog entry ending in `.yml` and one ending in `.yaml`
- run `chloggen update --config <config-from-collector-releases> --version <some-version>`
- check if all changelog entries were correctly deleted and added to `CHANGELOG.md`

### Expected Behavior

.yml and .yaml changelog entries should be treated the samew way

douglascamata:
Fixes #1079.

Adds support for reading and deleting entries that end up with `.yml` so that YAML files with this alternative extension are properly processed into the changelog.

### DIFF
--- a/.chloggen/chloggen-yml-support.yaml
+++ b/.chloggen/chloggen-yml-support.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. crosslink)
+component: chloggen
+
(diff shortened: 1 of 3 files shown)


mx-psi (approved):

douglascamata:
Thanks for the review, @MrAlias. Suggestions applied!

Issue #1243, opened 2025-09-22
GITHUB_REF env variable doesn't contain a PR number on push to main


jelly-afk:
### Component

issuegenerator

### Bug Description

In #1226 , I added a feature to include PR number in autogenerated issues which uses the GITHUB_REF env variable which assumes the format to be `refs/pull/<pr_number>/merge` which is true for workflows triggered by pull requests, but after going through the workflow file, 
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/7b460f275c4eac820c891a7324c6c2c201184cb6/.github/workflows/build-and-test.yml#L668-L669
the issue generator is used only on workflows triggerd by push on main, in which case the GITHUB_REF format is `refs/heads/main`.

This means issuegenerator will fail to extract pr number every time, so we should remove the pr field for now.

I apologize for this oversight.

### Steps To Reproduce

N/A

### Expected Behavior

N/A

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

PR #1244, opened 2025-09-22
[issuegenerator] remove pr section from issue body

jelly-afk:
fixes: #1243 

removed pr section from issue body as workflows triggered by push to main does not contain pr number as described in #1243 

Added shortSha function to only display first 7 characters of commit sha instead of full 40 characters.


mx-psi:
Thanks for taking a look, could you take a look at the fialing lint jobs?

PR #1357, opened 2025-10-29
[checkapi] manage struct name collisions

atoulme:
This is a problem picked up in the prometheusreceiver: the same struct name is used in the root of the module and under the targetallocator package.

The way we collect structs didn't account for the package name of the struct.

With this change, we will now store `Config` and `targetallocator.Config` as separate keys.


mx-psi (commented):
It's hard for me to understand what this PR is about at a quick glance, would you mind adding a PR description?

mx-psi (approved):

PR #1739, opened 2026-04-09
CI: Make a new job to run integration tests only on ubuntu

amishhaa:
Part of #1528.

- Add a new job test-integration that runs integration tests only on ubuntu.
- The integration tests would be detected using build tags.


dmathieu:
Now that we're only running grater tests when there are grater changes, I wonder if we need this right away. How about we start with the possibly slow docker tests without any build tag, and only make that change if it gets painful?
If we do this properly (with the interface we recommended), there shouldn't be too many slow docker tests.

amishhaa:
@dmathieu, we needed this because integration tests(eg:  Actually spinning a container and asserting the commands we ran actually worked) needs docker to be running, which is painful to set up for windows in the CI, so it is preferred to only run it for ubuntu that is why a job was created because test-race runs it on windows as well as ubuntu, if we want slow tests, we would only be able to mock. but yes we do not need this right away, we can get back to this if the CI for run PRs complain.

dmathieu:
Ah right, the docker on windows issue. Sorry I forgot that.

dmathieu (approved):

PR #1741, opened 2026-04-09
CI: AFFECTED_MODULES should exclude internal/tools and checkapi internals.

amishhaa:
We override variable ALL_GO_MOD_DIRS in Makefile, however in the construction of it there we exclude TOOLS_MOD_DIR and CHECKAPI_INTERNAL_MOD_DIRS, refer https://github.com/open-telemetry/opentelemetry-go-build-tools/blob/main/Makefile#L36,
however the script that currently generated affected packages for ci does not remove them which leads to unwanted dirs in AFFECTED_MODULES.

Skip the modules in TOOLS_MOD_DIR and CHECKAPI_INTERNAL_MOD_DIRS in CI when we pass AFFECTED_MODULES from detected-changes. Ad generalise CHECKAPI_INTERNAL_MOD_DIRS to INTERNAL_TEST_MOD_DIRS

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -191,4 +191,14 @@ jobs:
           echo "AFFECTED_FILES=$AFFECTED_FILES"
           chmod +x ./get_affected_pkgs.sh
           AFFECTED_MODULES=$(./get_affected_pkgs.sh $AFFECTED_FILES)
+
+          # Exclude tools and checkapi internal modules
+          TOOLS_MOD_DIR="./internal/tools"
+          CHECKAPI_INTERNAL_MOD_DIRS=$(find ./checkapi/internal -type f -name 'go.mod' -exec dirname {} \;)
+          
+          for exclude in $TOOLS_MOD_DIR $CHECKAPI_INTERNAL_MOD_DIRS; do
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -191,4 +191,14 @@ jobs:
           echo "AFFECTED_FILES=$AFFECTED_FILES"
           chmod +x ./get_affected_pkgs.sh
           AFFECTED_MODULES=$(./get_affected_pkgs.sh $AFFECTED_FILES)
+
+          # Exclude tools and checkapi internal modules
+          TOOLS_MOD_DIR="./internal/tools"
+          CHECKAPI_INTERNAL_MOD_DIRS=$(find ./checkapi/internal -type f -name 'go.mod' -exec dirname {} \;)
```
dmathieu: I would exclude anything that matches `internal/test*`.
We will likely end up having internal test modules in grater too.

dmathieu:
```suggestion
          CHECKAPI_INTERNAL_MOD_DIRS=$(find ./*/internal/test* -type f -name 'go.mod' -exec dirname {} \;)
```

dmathieu (approved):

PR #1758, opened 2026-04-14
grater: add basic interface to interact with docker.

amisha-chhajed:
Part of #1528.

- Add a new interface controller.
- Docker controller struct can create containers, hold references of all volumes and execute commands inside containers.
- ExecuteCommand method reports back the output.
- Mock Docker Controller implements interface with mock methods. Used for slow tests.


dmathieu:
This implementation sounds good. But this doesn't include an interface, only the docker struct.
We mentioned having (and using) a `Container` interface that would allow having multiple implementation structs (including a mock one), and only use actual docker when we really need it.

grater/internal/controller/controller.go
```diff
@@ -0,0 +1,16 @@
...
+
+// Package controller provides an interface for managing Docker containers and volumes.
+package controller
+
+import (
+    "github.com/docker/docker/api/types/container"
+)
+
+// Controller is an interface for managing Docker containers and volumes.
+type Controller interface {
```
dmathieu: IMHO, `Controller` is too generic. It could be many things. I would name this `Container`, even though it may not actually be a "container" in the docker sense.

amishhaa: Renamed the struct to Container, however the package controller cannot be renamed to container as it was causing a import conflict with docker/container.
Troublesome import pattern:
>
	"github.com/docker/docker/api/types/container"
	"go.opentelemetry.io/build-tools/grater/internal/container"

Now:
>
	"github.com/docker/docker/api/types/container"
	"go.opentelemetry.io/build-tools/grater/internal/controller"

dmathieu: You could import the docker container package with a specific name.

```
dockerContainer "github.com/docker/docker/api/types/container"
"go.opentelemetry.io/build-tools/grater/internal/container"
```

grater/internal/container/container.go
```diff
@@ -0,0 +1,19 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build integration
+// +build integration
```
dmathieu, line 5: Only tests should have the integration build. Otherwise. We can't use them at runtime.

grater/internal/dockercontroller/mockdockercontroller.go
```diff
@@ -0,0 +1,49 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// Package dockercontroller provides a mock implementation of the DockerController interface.
+package dockercontroller
```
dmathieu: This is a mock container, not specific to docker. How about having it in `grater/internal/mockcontainer`?

grater/internal/dockercontainer/dockercontroller.go
```diff
@@ -0,0 +1,135 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// Package dockercontroller provides a controller for managing Docker.
+package dockercontroller
```
dmathieu:
```suggestion
package dockercontainer
```

grater/internal/mockcontainer/mockdockercontroller.go
```diff
@@ -0,0 +1,49 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// Package mockcontainer provides a mock implementation of the DockerController interface.
```
dmathieu:
```suggestion
// Package mockcontainer provides a mock implementation of the Container interface.
```

grater/internal/mockcontainer/mockdockercontroller.go
```diff
@@ -0,0 +1,49 @@
...
+package mockcontainer
+
+import (
+	"github.com/moby/moby/client"
+
+	"go.opentelemetry.io/build-tools/grater/internal/container"
+)
+
+// MockDockerContainer is a mock implementation of the DockerContainer interface.
```
dmathieu:
```suggestion
// MockDockerContainer is a mock implementation of the Container interface.
```

grater/internal/container/container.go
```diff
@@ -0,0 +1,14 @@
...
+// Package container provides an interface for managing Docker containers and volumes.
+package container
+
+import "github.com/moby/moby/client"
+
+// Container defines the interface for managing Docker containers and volumes.
+type Container interface {
+	CreateVolume(volumeName string) (func(), error)
+	UseContainer(imageName string, volumeNames []string) (string, func(), error)
+	ExecuteCommand(containerID string, cmd []string) (string, client.ExecInspectResult, error)
```
dmathieu: This interface shouldn't have a dependency on moby, it should be generic (only one implementation will use moby).
Also, we can improve that later, but I'd return structs rather than `func` or `string`, to make things more generic as well.

dmathieu (commented):
We're getting there 😸

grater/internal/container/container.go
amishhaa: Instead of client.ExecInspectResult I have updated to return int of exitcode as the usecase of this was to eventually know the exitcode of commands only, as we already return out separately as string.

amishhaa: Yes, its better to make a generic struct response like, ExecuteCommandResponse, which we can extend and modify as we want in future, how about we make a follow up PR for this?

dmathieu (approved):

mx-psi (approved):

PR #1810, opened 2026-04-30
grater: Add support for response, config structs and copying host path dirs to containers.

amisha-chhajed:
Part of #1528.
- Add support for copying content of local paths with tar in containers to support local path dir copying into containers.
- As we are supporting the functionality to add dependents as local paths we need to copy the snapshot in a container,
so when we edit the go.mod to inject deps, changes stay isolated in the container.
- Add config structs to configure the inputs of any methods of the container.
- Add response structs for container methods.


grater/internal/container/config.go
```diff
@@ -0,0 +1,164 @@
...
+	return cfg.containerID
+}
+
+// Cmd returns the command to execute.
+func (cfg *ExecuteCommandConfig) Cmd() []string {
+	return cfg.cmd
+}
+
+// NewExecuteCommandConfig applies all the options to a returned ExecuteCommandConfig.
+func NewExecuteCommandConfig(options ...ExecuteCommandOption) ExecuteCommandConfig {
```
mx-psi, line 23: Nice use of functional options! :)

grater/internal/container/config.go
```diff
@@ -0,0 +1,164 @@
...
+	config := ExecuteCommandConfig{containerID: "", cmd: []string{}}
```
mx-psi: You can just do `config := ExecuteCommandConfig{}` or even `var config ExecuteCommandConfig` and rely on the [zero values](https://go.dev/ref/spec#The_zero_value) for the field types. This is usually how it's done in Go.

This applies on the other constructors like `NewCreateVolumeConfig` but just making one comment here

mx-psi (approved):
Just one small style comment, otherwise looks good to me!

grater/internal/container/config.go
amisha-chhajed: Thank you!!! :))

grater/internal/container/config.go
amisha-chhajed: Done, thank you for pointing me in that direction, great resource! :)

PR #1862, opened 2026-05-10
grater: add version to Module struct as replace needs version of modules.

amisha-chhajed:
Part of #1528.
- Modules have attribute version which we need in go mod edit -replace to specify the old module.In go.mod files we need the version of the previous module to replace it with a new one.
- In case of dependent being remote, there are multiple versions(branches) that we could clone, and users must be able to specify this.
- Add version field in module struct which is optional to fill in case of dependent, and needed for mainModule and injections for which we run go mod edit command.


mx-psi (approved):
Looks great!

PR #1866, opened 2026-05-12
grater: Add methods for executing basic repetitive commands.

amisha-chhajed:
Part of #1528.
- Runner uses a lot of repetitive commands which can be consolidated in a commands/ package to keep them separate and clean from the logic of runner itself.
- Separate commands so we can test their logic independently of runner.
- Add testdata/ with dependent module and modulePass and moduleFail which can be used in commands and runner to invoke different types of testing scenarios.

### DIFF
--- a/grater/internal/commands/commands.go
+++ b/grater/internal/commands/commands.go
@@ -12,7 +12,7 @@ import (
 )
 
 // GetModule downloads a module via the Go module proxy to the given path.
-func GetModule(ctx context.Context, c container.Container, useContainerResp container.UseContainerResponse, module module.Module, modulePath string) error {
+func GetModuleFromProxy(ctx context.Context, c container.Container, useContainerResp container.UseContainerResponse, module module.Module, modulePath string) error {
 	script := fmt.Sprintf(
 		`set -e
 	GOMODCACHE=$(mktemp -d)
--- a/grater/internal/commands/commands.go
+++ b/grater/internal/commands/commands.go
@@ -18,7 +18,7 @@ func ShallowClone(ctx context.Context, c container.Container, useContainerResp c
 	if branch != "" {
 		args = append(args, "--branch", branch)
 	}
-	args = append(args, "https://" + module.ModulePath + ".git", modulePath)
+	args = append(args, "https://"+module.ModulePath+".git", modulePath)
 
 	_, err := c.ExecuteCommand(ctx,
 		container.NewExecuteCommandConfig(
@@ -49,8 +49,8 @@ func CheckoutBranch(ctx context.Context, c container.Container, useContainerResp
 }
 
 // SetReplaceDirective adds a new replace directive in the go.mod file on the given path.
-func SetReplaceDirective(ctx context.Context, c container.Container, useContainerResp container.UseContainerResponse, old, new, modulePath string) error {
-	replace := fmt.Sprintf("%s=%s", old, new)
+func SetReplaceDirective(ctx context.Context, c container.Container, useContainerResp container.UseContainerResponse, oldRef, newRef, modulePath string) error {
+	replace := fmt.Sprintf("%s=%s", oldRef, newRef)
 
 	_, err := c.ExecuteCommand(ctx,
 		container.NewExecuteCommandConfig(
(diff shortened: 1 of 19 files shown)


grater/internal/commands/commands.go
```diff
@@ -0,0 +1,98 @@
...
+
+	"go.opentelemetry.io/build-tools/grater/internal/container"
+	"go.opentelemetry.io/build-tools/grater/internal/module"
+)
+
+// ShallowClone shallow clones a module inside a container to the given path.
+func ShallowClone(ctx context.Context, c container.Container, useContainerResp container.UseContainerResponse, module module.Module, branch, modulePath string) error {
+	args := []string{"git", "clone", "--depth=1"}
+	if branch != "" {
+		args = append(args, "--branch", branch)
```
mx-psi: We should put the branch name inside quotes and sanitize the input to avoid creating a situation where untrusted input leads to arbitrary command execution opportunities like in https://github.com/advisories/GHSA-8v8w-v8xg-79rf.

Looking at this more closely, I think it should be fine because we are not doing `sh -c` as below, but in any case it is a good practice

grater/internal/commands/commands.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+	args = append(args, "https://"+module.ModulePath+".git", modulePath)
```
mx-psi: (This can be left for later, wirting it down so we have it somewhere)

If the module path can be a generic Go module such as `go.opentelemetry.io/pdata` or `github.com/open-telemetry/opentelemetry-collector-contrib/processor/transformprocessor` this way of cloning won't work since, in the first case, this is not a VCS system and, in the second case, it refers to a sub-folder inside a VCS repository (see [here](https://pkg.go.dev/cmd/go#hdr-Remote_import_paths) for full details). 

The easiest way to support that would be downloading from the Go proxy, like in https://proxy.golang.org/github.com/open-telemetry/opentelemetry-collector-contrib/processor/transformprocessor/@v/v0.152.0.zip

It's unclear to what extent we want to support this and we don't need to do this now, we can save it as a task for later.

grater/internal/commands/commands_test.go
```diff
@@ -0,0 +1,120 @@
...
+        container.WithContainerID(useContainerResp.ContainerID),
+        container.WithCommand([]string{"git", "-C", modulePath, "rev-parse", "--is-shallow-repository"}),
+		),
+	)
+	require.NoError(t, err)
+	assert.Equal(t, "true", resp.Output)
+}
+
+func TestCheckoutBranch(t *testing.T) {
+	// TODO: Add remote testing functionality for checkout branch.
```
mx-psi: You can use a personal repository here since this repository does not have any secrets on CI. I will work on making sure that we use an OTel-controlled repository for testing later

grater/internal/commands/commands.go
```diff
@@ -0,0 +1,98 @@
...
+
+// SetReplaceDirective adds a new replace directive in the go.mod file on the given path.
+func SetReplaceDirective(ctx context.Context, c container.Container, useContainerResp container.UseContainerResponse, oldRef, newRef, modulePath string) error {
+	replace := fmt.Sprintf("%s=%s", oldRef, newRef)
+
+	_, err := c.ExecuteCommand(ctx,
+		container.NewExecuteCommandConfig(
+			container.WithContainerID(useContainerResp.ContainerID),
+			container.WithCommand([]string{
+				"sh", "-c", fmt.Sprintf(`cd %s && go mod edit -replace %s`, modulePath, replace),
```
mx-psi: To avoid having to do `cd %s` I think we could use `WorkingDir` which an option in `ExecClientOptions` on the Docker Golang package (I have not checked this closely so please check it yourself if you want to try this option, I am not sure it works)

grater/internal/commands/commands.go
amisha-chhajed: This is a great way to get remote modules, one way is to use go mod download, which automatically puts it inside go mod cache, copy it to our workingDir and delete go mod cache which is a simpler alternative to using curl commands to get the module from the archive, this abstracts the idea of cloning and links inside go mod download.

This also was able to catch a shortcoming that when we run our find script, responsible to give modules that depend on our module, it returns those dependent modules with versions, and version is a requirement for us to GetModule, and is simple to harden as a input from user with regex instead of relying on branches.

How about initially we support this way of getting modules to have a strong E2E model in place, then we will be able to much simply extend other ways of adding a module(for example: via a GitHub url)?

grater/internal/commands/commands_test.go
amisha-chhajed: currently CheckoutBranch method wouldn't be needed as it was specific in the case where we are giving mainModule locally and want to checkout between two versions, how about we first try to build it with module path and versions specific to go, then add different ways of having those modules inside the runner iteratively once we have a model that works for Go modules perfectly E2E? 

For example for now we have GetModuleFromProxy, similarly we can have a GetModuleFromUrl, which will be much clearer to extend once we have runner made E2E.

grater/internal/commands/commands.go
amisha-chhajed: Thank you, WorkingDir reduces the need for cd and makes everything cleaner!

grater/internal/commands/commands.go
amisha-chhajed: Thank you for pointing me to this, refactored to avoid using any user controlled inputs other than ModulePath and ModuleVersion which has some regex validation in place, we can further harden in a follow up PR

mx-psi (approved):

PR #1870, opened 2026-05-14
CI: fix coverage report by only reporting coverage for changed modules.

amisha-chhajed:
Part of #1528.

- Integration tests might be flaky as they rely on network calls, separate them from coverage report if changes are not made in those modules.


mx-psi (approved):
Let's try this out!

PR #1899, opened 2026-05-23
grater: Add Core Runner that runs tests of dependents of our main modules with specified replacements.

amishhhaaaa:
Part of #1528.
- Add Core runner.


grater/internal/module/module.go
```diff
@@ -29,6 +29,6 @@ func NewModule(modulePath string, moduleVersion string) *Module {
 
 // IsRemotePath checks if the module is a remote path.
 func (m *Module) IsRemotePath() bool {
-	match, _ := regexp.MatchString(`^github\.com/[^/]+/[^/]+`, m.ModulePath)
+	match, _ := regexp.MatchString(`^go\.`, m.ModulePath)
```
mx-psi: In principle, any URL could be a valid module, we could just use https://pkg.go.dev/net/url#Parse for now and then tighten this later

amisha-chhajed: Done. Added a generalized regex for valid urls as net url parse needs a prefix of http to parse the rest correctly or it falls back rest of text in the path field.

PR #1932, opened 2026-06-05
[githubgen/codeowners] fix unmaintained codeowners path

VihasMakwana:
We should trim path of unmaintained components, relative to `-folder` value. Similar to what we do for normal components. 
CI has been failing because this bug https://github.com/open-telemetry/opentelemetry-collector-contrib/actions/runs/26904407964/job/79365279559?pr=48856


githubgen/codeowners_test.go
```diff
@@ -310,6 +311,107 @@ func Test_codeownersGenerator_Generate(t *testing.T) {
...
+							Stability: map[string][]string{
+								unmaintainedStatus: {""},
+							},
+						},
+					},
+				},
+			}
+
+			require.NoError(t, cg.Generate(data))
+			require.Containsf(t, string(capturedContent), tt.wantContains, "generated CODEOWNERS does not contain %q", tt.wantContains)
```
MovieStoreGuy, line 412: Should this uses equals instead of contains?

VihasMakwana: I think we just want to verify if a given component is present in generated content. `require.Containsf` works fine in this case.

MovieStoreGuy (approved):

PR #2034, opened 2026-07-18
[chore] Switch to mdlint cli2

thompson-tomo:
Switch to a mantained markdownlint docker image/gh-action.

Also added a make action like other go repo's


mx-psi (commented):
Looks good! Are we already using this action in any other OTel repo?

thompson-tomo:
Yes we are. One such example https://github.com/open-telemetry/semantic-conventions/blob/main/.github/workflows/checks.yml

mx-psi (approved):

PR #2086, opened 2026-08-03
[chore] pin securego/gosec to v2.28.0 instead of master

opentelemetrybot:
`.github/workflows/gosec.yml` pins `securego/gosec` by digest but comments the ref as `# master`. Renovate keeps that digest chasing gosec's `master` branch, so every time `master` advances the comment no longer describes what the digest points at, and zizmor's [`ref-version-mismatch`](https://docs.zizmor.sh/audits/#ref-version-mismatch) fires:

```
warning[ref-version-mismatch]: action's hash pin has mismatched or missing version comment
  --> .github/workflows/gosec.yml:28:73
   |
28 |         uses: securego/gosec@45b083a0cb42119d61b1e0e364d8c62c68f6f5cd # master
   |                                                                         ^^^^^^ points to commit 5e72efc19d97
```

This recurs indefinitely — it will re-fire on the next `master` commit — so bumping the digest again is not a fix.

The `@master` ref dates back to the original workflow in #7, which copied the example from gosec's README. gosec documents `@v2` alongside `@master`, so there is no reason this repo needs to track unreleased code.

This pins the latest release instead, `v2.28.0` (`9e75c0576c9878035d4221392108d458abe10fc3`). Because `renovate.json` already extends `helpers:pinGitHubActionDigestsToSemver`, Renovate will keep this current against real releases and the version comment will stay accurate.

Verified with zizmor v1.27.0 (online audits enabled): the finding reproduces on `main` and is resolved on this branch.

Follow-up to #2071. Part of open-telemetry/sig-security#293.


MrAlias (approved):

MrAlias:
@trask is there automation to rebase this?

trask:
no, but I just asked it to

PR #2182, opened 2026-09-03
Pin changelog link checker dependencies

MrAlias:
- install markdown-link-check from a committed npm lockfile
- invoke the lockfile-backed local executable in the changelog workflow


dmathieu (commented):
Shouldn't we update the renovate config to look for upgrades in this package.json?

codeboten (commented):
The changelog now uses the common shared workflow in https://github.com/open-telemetry/shared-workflows/tree/main/chloggen, i think this can be closed

PR #2212, opened 2026-09-18
feat(chloggen): group changelog entries by component

### LINKED ISSUE #971: chloggen: add support for grouping changes by component
Users don't necessarily care about changes to all components in a changelog. For example, users of the contrib collector are likely to use only a handful of receivers and exporters, rather than all of them.

To avoid users having to search through the changelog for entries relevant to them, it would be helpful to group changes by component.

See https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/40593

r0hansaxena:
## Description:

Fixes #971

Adds a `--group-by-component` flag to `chloggen update`. When set, entries for the same component are listed under a single bullet within each change type, so users can skip the components they don't use. Without the flag the output is unchanged.

Useful for contrib, which can pass the flag in its `chlog-update` target.

### Verification
- added `TestGroups` and a `group_by_component` case in `TestUpdate`
- `go test ./...`, `golangci-lint run`

## Checklist
- [x] I have signed the [CLA](https://docs.linuxfoundation.org/lfx/easycla/contributors/getting-started).
- [x] Added changelog entry
- [x] Tests passed


r0hansaxena:
@dmathieu PTAL

dmathieu:
I think this should be an option, not applied everywhere. It's useful for contrib, but not so much for core.

r0hansaxena:
> I think this should be an option, not applied everywhere. It's useful for contrib, but not so much for core.

do you mean a config setting, something like `group_by_component: true`? or would it be enough to keep the default template as it is and have contrib use `groupByComponent` in its own summary template?

dmathieu:
We definitely don't want an option within the template. This isn't a decision that should be made per changelog entry.
A CLI option, so the call to chloggen on release decides how data is grouped?

r0hansaxena:
> We definitely don't want an option within the template. This isn't a decision that should be made per changelog entry. A CLI option, so the call to chloggen on release decides how data is grouped?

moved it behind a `--group-by-component` flag on `chloggen update`. without the flag the output is the same as before

dmathieu:
cc @open-telemetry/collector-contrib-approvers is this indeed something that would make your changelog better?

atoulme:
Can you post a gist of how that would look like? you group by component but by category first? If so, that sounds less useful.

I'd love to have each component have its own changelog, if you're so inclined to make it happen. In which case each component would have its own changelog file under their own folder.

ArthurSens:
Doesn't chloggen accept custom Go templates? Isn't "group components together" just a templating issue?

r0hansaxena:
> Can you post a gist of how that would look like? you group by component but by category first? If so, that sounds less useful.
> 
> I'd love to have each component have its own changelog, if you're so inclined to make it happen. In which case each component would have its own changelog file under their own folder.

```
- `receiver/foo`:
  - add X (#1)
  - add Y (#2)
- `receiver/bar`: add Z (#3)
```

the change type headings don't change, entries for the same component just sit under one bullet instead of repeating the name.

for per-component changelogs, entries would need to map to their folder somehow, `receiver/filelog` vs `receiver/filelogreceiver`. should I open a separate issue?

r0hansaxena:
> Doesn't chloggen accept custom Go templates? Isn't "group components together" just a templating issue?

it does, but the template only gets a flat list, so you'd have to track the previous component as you range over it. the flag just moves that decision to release time.

PR #2215, opened 2026-09-20
feat(multimod): allow re-running tag on an already tagged commit

### LINKED ISSUE #205: Make it possible to run the same `multimod tag` command multiple times
It would be nice to be able to do a dry run of `git push` based on the output of `multimod tag --print-tags`, but it'll make the second run of `multimod tag` fail on the same commit because tags are already created. `multimod tag` potentially can be updated to check if the tags are already created on the specified commit just return a similar output instead of failing.

r0hansaxena:
## Description:

Fixes #205

Running `multimod tag` again on the same commit fails because the tags are already there, so you can't use `--print-tags` to dry run the push and then run it a second time.

`newTagger` now checks if the tags are already on that commit and skips tagging if they are, the tags still get printed. Still errors if they're on a different commit, or if only some of them exist.

### Verification
- updated the already-exists case in `TestTagAllModules`, added one for tags on a different commit
- `go test ./...`, `golangci-lint run`

## Checklist
- [x] I have signed the [CLA](https://docs.linuxfoundation.org/lfx/easycla/contributors/getting-started).
- [x] Added changelog entry
- [x] Tests passed


r0hansaxena:
@codeboten PTAL

MrAlias:
I see not allowing duplicate tagging as a feature, not a bug. Re-tagging can be catastrophic for Go versioning. What reasoning motivates adding such a dangerous option?

r0hansaxena:
> I see not allowing duplicate tagging as a feature, not a bug. Re-tagging can be catastrophic for Go versioning. What reasoning motivates adding such a dangerous option?

it doesn't re-tag anything. if the tags already point at that commit it skips creating them and prints the list and it still errors if any are on a different commit or only some exist.

the case I had in mind was re-running `multimod tag --print-tags` on the same commit to dry-run the push, which currently errors out.

